PR #23, opened 2023-12-20
Change firebase lib new

nevogithub:
Replacing firebase lib to a server specific one.

### DIFF
--- a/cv_next/server/api/cvs.ts
+++ b/cv_next/server/api/cvs.ts
@@ -8,16 +8,10 @@ import {
   DocumentData,
   Query,
   QueryDocumentSnapshot,
-  QueryFieldFilterConstraint,
-  collection,
-  limit,
-  orderBy,
-  query,
-  startAfter,
-  where,
-  documentId,
   CollectionReference,
-} from "firebase/firestore"
+  FieldPath,
+  Filter
+} from "firebase-admin/firestore"
 import Definitions from "../base/definitions"
 import FirebaseHelper from "./firebaseHelper"
 import { DbOperationResult, ErrorReasons, RateLimitError } from "./utils"
@@ -51,22 +45,21 @@ export function documentSnapshotToCV(
  * Retrieves paginated CVs based on a query filter and optionally filters out deleted CVs.
  * If last document id is given, retreive the "page" that starts after it.
  *
- * @param {QueryFieldFilterConstraint} queryFilter - Optional query filter to apply to the CVs.
+ * @param {Filter} queryFilter - Optional query filter to apply to the CVs.
  * @param {boolean} filterOutDeleted - Optional flag to filter out deleted CVs. Defaults to true.
  * @param {string?} lastId - Optional last cv document id to start the query after.
  * @return {Promise<CvModel[] | null>} A promise that resolves to an array of CV models or null if an error occurred.
  */
 export async function getPaginatedCvsByQueryFilter(
-  queryFilter?: QueryFieldFilterConstraint,
+  queryFilter?: Filter,
   filterOutDeleted: boolean = true,
   lastId?: string
 ): Promise<CvModel[] | null> {
-  let pageQuery: Query<DocumentData>
-  let collectionRef = collection(
-    FirebaseHelper.getFirestoreInstance(),
+  let pageQuery: Query<DocumentData>;
+  let collectionRef = FirebaseHelper.getFirestoreInstance().collection(
     CvModel.CollectionName
-  )
-  let filterDeletedQuery = where("deleted", "==", !filterOutDeleted);
+  );
+  let filterDeletedQuery = Filter.where("deleted", "==", !filterOutDeleted);
   
   if (lastId) {
     const lastDoc = await getCvById(lastId, collectionRef, false);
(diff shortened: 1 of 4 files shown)


cv_next/server/api/firebaseHelper.ts
```diff
@@ -1,25 +1,25 @@
 import 'server-only'
 
-import { FirebaseApp, initializeApp } from "firebase/app";
+import { App, initializeApp, getApp, getApps} from "firebase-admin/app";
+import { createRequire } from 'module'
+const require = createRequire(import.meta.url)
+const cvSdk = require("@/cv-sdk.json");
```
YechiamTK, line 6: A. What is "cvSdk"?
B. Why import using "require"?

cv_next/server/api/firebaseHelper.ts
```diff
@@ -1,25 +1,25 @@
...
   Query,
-} from "firebase/firestore";
+  getFirestore,
+  Firestore
+} from "firebase-admin/firestore";
 import Definitions from "@/server/base/definitions";
 import OperationBucket from "@/server/api/operationBucket";
 import MyLogger from "@/server/base/logger";
 import { RateLimitError, enumToStringMap, ErrorReasons } from "./utils";
+import { fromJSON } from 'postcss';
```
YechiamTK, line 19: Is this used anywhere?

cv_next/server/api/firebaseHelper.ts
```diff
@@ -44,15 +44,18 @@ export default class FirebaseHelper {
      * Method initiates the firebase app if not initiated yet
      * @returns FirebaseApp instance
      */
-    public static getFirebaseInstance(): FirebaseApp {
+    public static getFirebaseInstance(): App {
+      
       if (
-        FirebaseHelper.firebaseAppInstance === null ||
-        FirebaseHelper.firebaseAppInstance === undefined
+        (FirebaseHelper.firebaseAppInstance === null ||
```
YechiamTK, line 50: Unnecessary extra brackets. Also `!` is preferable to `null` or `undefined` checks, unless explicitly needed.

cv_next/server/api/firebaseHelper.ts
```diff
@@ -44,15 +44,18 @@ export default class FirebaseHelper {
...
+        FirebaseHelper.firebaseAppInstance === undefined) 
       ) {
-        FirebaseHelper.firebaseAppInstance = initializeApp(
-          Definitions.FIREBASE_CONFIG
-        );
+        FirebaseHelper.firebaseAppInstance = (getApps().length === 0) ? initializeApp({
+          credential: credential.cert(cvSdk)
+        }, "cv-next") :
```
YechiamTK, line 55: Please use a constant for `"cv-next"` - perhaps change the project name in `package.json` to its proper name, and import it from there?
If you don't want to create a hard relationship between the project name and the firebase project, you can also save this name as a global constant instead (in a "constants" folder)

cv_next/server/api/cvs.ts
```diff
@@ -75,37 +68,32 @@ export async function getPaginatedCvsByQueryFilter(
...
-        startAfter(lastDoc),
-        filterDeletedQuery
-      );
+    ? 
+    collectionRef
+    .orderBy("uploadDate", "desc")
+    .limit(queryLimit)
+    .where(filterDeletedQuery)
+    .where(queryFilter)
+    .startAfter(lastDoc)
```
YechiamTK, line 77: If we're using a package that let's us use "skip", could you please refactor this so that instead of passing a `lastDoc` you'll receive a page number and skip according to page number * pre-set amount of cvs in page (can also receive this, with a default amount of let's say 10)

cv_next/server/api/users.ts
```diff
@@ -97,24 +95,24 @@ export default class UsersApi {
...
       );
+      
       let querySnapshot: QuerySnapshot;
-      let activeWhere = where("active", "==", true);
+      let activeWhere = collectionRef.where("active", "==", true);
       if (w !== undefined) {
         const q = filterOutInactive
-          ? query(collectionRef, w, activeWhere)
-          : query(collectionRef, w);
+          ? collectionRef.where(w).where(activeWhere)
```
YechiamTK, line 110: Can the new package allow for editing the query after initialization?
If so, a better code will be:
`let q = collectionRef.where(w);
if (filterOutInactive) {
    q = q.wheree(activeWhere); //or something similar
}`

YechiamTK: Also as `filterOutInactive` check is probably common, if using this you can refactor to a separate function that'd be helpful.

YechiamTK (changes requested):

cv_next/server/api/firebaseHelper.ts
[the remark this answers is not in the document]
nevogithub: This is some weird workaround. Cv-sdk is a rename for the json file for authenticating the app. I needed to import it directly to the program, and in my tests, "require" did the job, unlike "import". If this is crucial, I'll check again.

cv_next/server/api/cvs.ts
nevogithub: Good idea. Thanks. I'll look into that.

cv_next/server/api/firebaseHelper.ts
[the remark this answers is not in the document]
YechiamTK: You mean the authentication settings that are found in `firebase.config.js`? Why would you rename it?
Also yes, please try to use import, consistency is important. If there's an actual problem that prevents you from using `import`, let me know and use `require` instead, but I don't see a reason why that'd happen.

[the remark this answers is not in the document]
nevogithub: The authorizing key for admin is different from the regular. But it's ok, I'll change the definitions file so it'll fit with the original route of process env variables.

cv_next/server/api/cvs.ts
nevogithub: Yeah well I looked into that - pretty sure that StartAt\After use the doc id, like the original implementation. Offset, on the other hand, is recommended against, in multiple sources, such as this one:
https://firebeast.dev/tips/do-not-use-offset
Apparently, "offset" severely increases IO. In the meantime, I'll keep the original implementation

cv_next/server/api/firebaseHelper.ts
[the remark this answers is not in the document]
YechiamTK: Oh man, this is the main reason I wanted this package, precisely because this is wrong in our case: "Besides being snappy fast, it will provide better UX as the same page will always produce the same results, unlike pagination by offset where items constantly shift." - for changing data (e.g. cv can be deleted), fetching by last doc id will _not_ always produce the same result. This is way for UX reasons, offset is better.
But if firebase doesn't allow us to query offset efficiently, I'll just keep on hating firebase and we'll keep the original implementation.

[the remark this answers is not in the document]
nevogithub: I apologize for not catching this sooner.. ooof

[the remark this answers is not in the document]
nevogithub: Oh and another thing my dumbass (pardon my French) found out last night - firestore security rules don't apply on Firebase Admin.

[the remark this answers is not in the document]
YechiamTK: Wait. Are we using this feature? Does that mean we can't use Admin?
Why can't they make a proper package ffs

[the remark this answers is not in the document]
nevogithub: TBH after all this time, and thinking about how we'd like to authenticate users and have security rules that matter, I've come to the conclusion that we should backtrack on replacing the firebase library. I apologize for making quite a mess, that didn't really prove itself in any way. Is there anything I could do now?

PR #29, opened 2023-12-29
Manage user states

Aradon267:
This branch includes user authentication using Gmail, user page access control that is based on a role(on the feed page as this was the only page available), a "access denied" page, and an example on getting the user's data


Aradon267:
closing this as this is an old pr

PR #36, opened 2024-01-28
Supabase migration

avnogy:


cv_next/server/api/cvs.ts
```diff
@@ -1,277 +1,165 @@
...
-): Promise<CvModel[] | null> {
-  let res = getPaginatedCvsByQueryFilter(undefined, filterOutDeleted, lastId)
-  MyLogger.logDebug(
-    `requested all cvs(pagination) - last cv id: ${lastId}`
-  )
-  return res
+  page?: number
+): Promise<PaginatedCvsModel | null> {
+  try {
+    page = page ?? Definitions.DEFAULT_PAGINATION_FIRST_PAGE_NUMBER
```
YechiamTK, line 114: Instead of doing this you can default-assign page to the default (i.e. in function definition). Also accordingly you can call the function without providing `undefined` but simply pass nothing.

cv_next/server/api/cvs.ts
```diff
@@ -1,277 +1,165 @@
...
-  CollectionReference,
-} from "firebase/firestore"
-import Definitions from "../base/definitions"
-import FirebaseHelper from "./firebaseHelper"
-import { DbOperationResult, ErrorReasons, RateLimitError } from "./utils"
+import Definitions from "../../lib/definitions"
+import SupabaseHelper from "./supabaseHelper"
+import { PostgrestError } from "@supabase/supabase-js"
 
 const queryLimit: number = Definitions.CVS_PER_PAGE
```
YechiamTK, line 9: I don't think this should (stay?) be defined globally in this file; just use the Definitions constant in the used functions. Perhaps add a query limit parameter to optionally bypass the default

cv_next/server/api/cvs.ts
```diff
@@ -1,277 +1,165 @@
...
+    let query = supabase
+      .from("cvs")
+      .select("*")
+      .range(from, to - 1)
+
+    if (filterOutDeleted) {
+      query = query.eq("deleted", false)
+    }
+
+    const { data: cvs, error } = await query
```
YechiamTK, line 129: Please don't forget semicolons. Run the prettier on all files (we should add this as a pre-commit hook)

cv_next/app/global.d.ts
```diff
@@ -0,0 +1,7 @@
+import { Database as DB } from "@/types/database.types"
+
+declare global {
+  type Database = DB
+  type CvModel = DB["public"]["Tables"]["cvs"]["Row"]
```
YechiamTK, line 5: I'm not a 100% sure models should be defined in a global d.ts file. Especially not one like PaginatedCvsModel. These should be in a models file/folder

cv_next/server/api/cvs.ts
avnogy: ill look into that

cv_next/types/database.types.ts
```diff
@@ -0,0 +1,182 @@
+export type Json =
+  | string
```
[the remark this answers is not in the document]
avnogy: if its in here its already exposed :) 
we can make it Sever-Only just for the sake of full division though.

cv_next/app/global.d.ts
avnogy: if its here its automatically picked up by eslint and vscode, and you dont have to import it every time. i saw it in an example somewhere and it seemed to me like the right place but we can talk about it if you're unconvinced

avnogy: resolving for now

avnogy:
note to self: fix revisiting feed page lets load the same cvs again

cv_next/app/global.d.ts
YechiamTK: Don't have to import it doesn't necessarily mean good practice. For me it means I don't know where it's defined..

cv_next/server/api/cvs.ts
YechiamTK: Then change it :D

cv_next/app/feed/components/feed.tsx
```diff
@@ -57,14 +55,15 @@ export default function Feed({
 
   const fetchCvsCallback = useCallback(async () => {
     if (loadMore) {
-      const response = await fetchCvsForFeed({ page: page.current + 1 })
+      var nextPage = page.current + 1
```
YechiamTK, line 58: Do not var

cv_next/app/feed/page.tsx
```diff
@@ -1,14 +1,13 @@
 import { Suspense } from "react"
 import { fetchCvsForFeed } from "@/app/actions/fetchCvs"
 import Feed from "./components/feed"
+import Definitions from "@/lib/definitions"
```
YechiamTK, line 4: Unused import?

cv_next/server/api/cvs.ts
avnogy: is there a reason you want semicolons? 90% of our code is without them

cv_next/app/feed/components/CVItemRSC.tsx
```diff
@@ -1,44 +1,53 @@
-import Image from "next/image"
+import Image, { StaticImageData } from "next/image"
 import Link from "next/link"
 import generateImageUrl from "@/helpers/imageURLHelper"
 import Categories from "@/types/models/categories"
-import { ClientCvModel } from "@/types/models/cv"
+import { useState } from "react"
```
YechiamTK, line 5: Note that this is supposed to be a static, server side component. Using hooks is not allowed in server components (and I'm pretty sure you'd get at least a warning about it).
Render a placeholder image here (that  "blank" img), and move the actual stateful render down to the CvItem.tsx file

cv_next/app/feed/components/CVItemRSC.tsx
```diff
@@ -1,44 +1,53 @@
...
+import blankCvImg from "../../../public/images/blackCv.png";
```
YechiamTK, line 6: Fix import stylel (should be fixed with prettier)

cv_next/app/feed/components/CVItemRSC.tsx
```diff
@@ -1,44 +1,53 @@
...
 
   return (
     <>
       <Image
         width={500}
         height={500 * 1.4142}
         className="w-full rounded-lg p-2"
-        src={imageUrl}
+        src={imgSrc}
+        onError={handleImgError}
```
YechiamTK, line 32: Note that you can use `placeholder` property, perhaps will be better

cv_next/app/feed/components/CVItemRSC.tsx
```diff
@@ -1,44 +1,53 @@
...
             </div>
-            <p className="inline text-xs text-neutral-700/75">
+            <p className="inline text-xs text-neutral-400">
               {formattedDate}
             </p>
           </div>
           <Link
             href={categoryLink}
-            className="absolute bottom-0 right-0 mx-4 mb-2.5 rounded-full bg-neutral-200 px-3 py-1 text-sm font-semibold text-neutral-700 hover:underline"
+            className="absolute bottom-0 right-0 mx-4 mb-2.5 rounded-full bg-gray-700 px-3 py-1 text-sm font-semibold text-white hover:underline hover:bg-gray-400"
```
YechiamTK, line 48: General note to everything style-related: I don't know how exactly you're handling it now, but I think it'd be best if @PeroG28 will make a file with the necessary colors - optimally we'd have a fixed schema of colors that we know to use throughout the app.

cv_next/app/feed/components/feed.tsx
```diff
@@ -1,82 +1,130 @@
...
     if (loadMore) {
-      const fetchedCvs = await fetchCvs({lastId: lastCvRef.current});
-      
-      if (fetchedCvs && fetchedCvs?.length > 0) {
-        if (fetchedCvs[fetchedCvs.length - 1]?.id === lastCvRef.current) {
-          setLoadMore(false);
-        } else {
-          setCvs((prevCvs) => [ ...(prevCvs?.length ? prevCvs : []), ...fetchedCvs ]);
-        }
+      const nextPage = page.current + 1
```
YechiamTK, line 62: Run prettier. @Aradon267 do we have pre-commit hook for prettier/es-lint?

cv_next/app/feed/components/feed.tsx
```diff
@@ -1,82 +1,130 @@
...
+      },
+    })
+    setCvs([])
+    page.current = Definitions.PAGINATION_INIT_PAGE_NUMBER
+    setLoadMore(true)
+  }
+
+  const onFilterChange = (filters: filterObj) => {
+    setFilters(filters)
+    forceReload()
```
YechiamTK, line 101: Small potential for race with state. Probably needs testing that we currently don't have (do we have filters?), but the `forceReload` function will get called without you knowing if the filters state has already changed.

cv_next/app/feed/components/filterPanel.tsx
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useState } from "react"
+import { SearchBox } from "../../../components/ui/filters/searchbar"
+import { DropdownInput } from "../../../components/ui/filters/valueSelect"
+import Categories from "@/types/models/categories"
+import { ModeSwitch } from "@/components/ui/filters/modeSwitch"
+
+export interface filterObj {
```
YechiamTK, line 7: change name..

cv_next/app/feed/components/filterPanel.tsx
```diff
@@ -0,0 +1,46 @@
...
+    searchValue: string
+    categoryId: number | null
+}
+
+export const MODES:[string, string] = ["review", "published"]
```
YechiamTK, line 12: Not the best way to define a constant. You'd either want an enum here, or a simple:
`export const MODES = {'Review': 0, 'Published': 1};`
enum is probably the way to go:
`export enum MODES {'Review', 'Published'};`

YechiamTK: Also this might be better to be in Definitions because it'll probably be used elsewhere as well

cv_next/app/feed/components/filterPanel.tsx
```diff
@@ -0,0 +1,46 @@
...
+    useEffect(() => {
+        onChange({
+            categoryId: categoryId,
+            searchValue: searchValue
+        })
+    }, [searchValue, categoryId])
+
+    return (
+        <div className="mx-10 my-[10px] flex flex-row justify-between items-center gap-2">
+            <SearchBox placeHolder=" input text to search" value={searchValue} onChange={(value: string) => {setSearchValue(value)}}></SearchBox>
```
YechiamTK, line 32: Tip: for all the `onChange`s, you're probably fine with putting the setState function itself (e.g. `onChange={setSearchValue}`)

YechiamTK: Also can this be stored in the component logic and not the render code? `{Object.keys(Categories.category).map(key=> parseInt(key)).filter(key => !isNaN(key))}`
Might wanna memoize it at some point, so it'd be useful to keep it separated anyway (originally this is a code-style suggestion, but it's also a utilization suggestion).

cv_next/app/feed/page.tsx
```diff
@@ -1,15 +1,11 @@
-import { Suspense } from "react";
-import { fetchCvs } from "@/app/actions/fetchCvs";
-import Feed from "./components/feed";
+import { Suspense } from "react"
+import Feed from "./components/feed"
 
 export default async function Page() {
```
YechiamTK, line 4: We don't want to initial-fetch on server?

cv_next/app/login/page.tsx
```diff
@@ -0,0 +1,19 @@
...
+export default function Page() {
+  const supabase = useSupabase()
+
+  return (
+    <Auth
+      supabaseClient={supabase}
+      appearance={{ theme: ThemeSupa }}
+      theme="dark"
+      providers={["google"]}
+      redirectTo="http://localhost:3000/auth/callback"
```
YechiamTK, line 15: Don't merge with localhost urls

YechiamTK: If you want, the only place localhost urls are allowed is in a development env file. You're more than welcome to add it and add a run config with the dev env.

cv_next/app/logout/page.tsx
```diff
@@ -0,0 +1,16 @@
+"use client"
+import { redirect } from "next/navigation"
+import { Button } from "@/components/ui/button"
+import { useSupabase } from "@/hooks/supabaseHooks"
+export default function Page() {
+  const supabase = useSupabase()
+  supabase.auth.signOut()
+
+  redirect("/")
```
YechiamTK, line 9: I remember that I already commented on this? Why create a component with render code if you force-redirect on first render anyway?

cv_next/components/layout/navbar.tsx
```diff
@@ -15,98 +18,111 @@ const navLinks = [
...
 
   const handleClick = async () => {
-    setNavbar(false)
+    setIsNavbarMenuOpen(false)
   }
 
   useEffect(() => {
-    if (navbar) {
+    if (isSettignsOpen) {
       document.body.style.overflow = "hidden"
```
YechiamTK, line 32: Ok why are we accessing DOM directly here? Is there a real reason? @Aradon267 please look into this

cv_next/components/layout/navbar.tsx
```diff
@@ -15,98 +18,111 @@ const navLinks = [
...
-                    <path
-                      strokeLinecap="round"
-                      strokeLinejoin="round"
-                      d="M4 6h16M4 12h16M4 18h16"
-                    />
-                  </svg>
-                )}
-              </button>
-              <ModeToggle />
+  return <>
```
YechiamTK, line 38: This is a big and messy div hell. Please refactor this component and separate it into smaller ones.

cv_next/components/layout/popup.tsx
```diff
@@ -0,0 +1,40 @@
...
+    path: "/logout",
+  },
+]
+
+interface PopupProps {
+  closeCb: () => void;
+}
+
+export default function Popup({ closeCb }: PopupProps) {
+  return <div className="fixed top-0 w-full left-0 h-full flex justify-center items-center">
```
YechiamTK, line 23: Why not use HTML's dialog? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
Also looks like there's no click-outside-to-close functionality, see a solution reference for example here: https://stackoverflow.com/a/57463812

cv_next/components/ui/filters/modeSwitch.tsx
```diff
@@ -0,0 +1,26 @@
+
```
YechiamTK, line 1: The entire `\ui\filters` folder should probably be moved to `\feed\components`. Unless they are used elsewhere as well.

cv_next/components/ui/filters/modeSwitch.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const ModeSwitch = ({
+    values,
+    value,
+    onChange,
+  }: {
+    value: string,
+    values: [string, string],
+    onChange: (newValue: string) => void,
+  }) => {
+    const isRightSide = value === values[1]
```
YechiamTK, line 11: What is happening here?
Please talk to me about this component

cv_next/hooks/supabaseHooks.ts
```diff
@@ -0,0 +1,28 @@
+import { useContext } from "react"
```
YechiamTK, line 1: if it's under `/hooks` folder, I don't think we need to name the file `Hooks` :)

cv_next/lib/utils.ts
```diff
@@ -1,27 +1,27 @@
 import { type ClassValue, clsx } from "clsx"
```
YechiamTK, line 1: This file changes means there are conflicting prettier configs!

cv_next/middleware.ts
```diff
@@ -0,0 +1,73 @@
+import { createServerClient, type CookieOptions } from "@supabase/ssr"
```
YechiamTK, line 1: Add a "use server"?

cv_next/prettier.config.js
```diff
@@ -1,7 +1,7 @@
-// prettier.config.js or .prettierrc.js
 module.exports = {
-  trailingComma: 'es5',
+  trailingComma: "es5",
   tabWidth: 2,
   semi: false,
```
YechiamTK, line 4: Since when do we have semi: false. Why.

cv_next/types/database.types.ts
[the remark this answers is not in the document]
YechiamTK: @Aradon267 please check this.

cv_next/app/feed/components/CVItemRSC.tsx
Aradon267: Specifically, this file is updated in a different branch that updates the preview source to be supabase so this is outdated

cv_next/app/feed/components/CVItemRSC.tsx
Aradon267: I'll update that but still, this file is outdated

cv_next/app/feed/components/CVItemRSC.tsx
Aradon267: Instead of onError?

cv_next/app/feed/components/feed.tsx
Aradon267: I'll check
BTW ran it but I didn't see these lines

cv_next/app/feed/page.tsx
Aradon267: The fetching done at the feed itself(not the CV itemRSC) only gets the CV entry(the link, ID, author, etc) while the preview is at thee RSC

cv_next/app/login/page.tsx
Aradon267: So what should I put there for the time being? Should I put the URL as a process env?

cv_next/app/logout/page.tsx
Aradon267: This was made when the Signout was as a "link" in the navbar.
Now this page is accessed from a button in the user profile component which again, redirects to this as only the supabase things are needed. Should the functionality also be a "link" that just doesn't redirect?

cv_next/components/layout/navbar.tsx
Aradon267: Don't know the difference that it makes(tried removing it locally)... @PeroG28 any idea If this is necessary?

cv_next/lib/utils.ts
Aradon267: Maybe that's tee semi false?

cv_next/app/feed/components/feed.tsx
Aradon267: According to what I see we don't have it
Wanted to add it but seems that for prettier this is [deprecated](https://github.com/pre-commit/mirrors-prettier#archived). will add for eslint

cv_next/app/login/page.tsx
YechiamTK: Exactky, urls should be constants and urls that depends on env should be process env constants

cv_next/app/feed/components/feed.tsx
YechiamTK: Check this: https://prettier.io/docs/en/precommit.html (especially first option)

cv_next/app/feed/components/CVItemRSC.tsx
YechiamTK: Yeah because in the current behavior there's an implicit use of client-side behavior (states). Basically all the `handleImgError` functionality is probably not needed and can be handled in Next-native behavior that'd support static render.

cv_next/app/feed/components/CVItemRSC.tsx
YechiamTK: BTW - outdated files that are updated in different branches - we'll merge this (big) branch and the other branches would resolve conflicts for themselves (and not, for example, merging a side-branch to this big branch).

cv_next/server/api/comments.ts
```diff
@@ -1,19 +1,11 @@
...
-import CommentModel from "@/types//models/comment";
-import MyLogger from "@/server/base/logger";
-import { DbOperationResult, ErrorReasons, RateLimitError } from "./utils";
-import FirebaseHelper from './firebaseHelper'
+import Helper from "@/server/base/helper"
+import CommentModel from "@/types//models/comment"
+import MyLogger from "@/server/base/logger"
+import { DbOperationResult, ErrorReasons, RateLimitError } from "./utils"
 
+//TODO: implement CommentsApi with supabase
```
YechiamTK, line 8: Is this TODO still relevant?

cv_next/server/api/comments.ts
```diff
@@ -1,19 +1,11 @@
-import 'server-only'
+import "server-only"
```
YechiamTK, line 1: Note that the changes in this file are probably just mis-configured prettier changes

cv_next/server/api/cvs.ts
```diff
@@ -1,277 +1,179 @@
...
-    )
-    let filterDeletedQuery = where("deleted", "==", !filterOutDeleted)
-    const q =
-      queryFilter !== undefined
-        ? query(collectionRef, queryFilter, filterDeletedQuery)
-        : query(collectionRef, filterDeletedQuery)
-    const querySnapshot = await FirebaseHelper.myGetDocs(q)
-    const matchingDocuments: CvModel[] = []
+    const supabase = SupabaseHelper.getSupabaseInstance()
+    let query = supabase.from("cvs").select("*").eq("user_id", userId)
```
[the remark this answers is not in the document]
YechiamTK: I commented here but it's relevant to all strings in this file

cv_next/server/api/cvs.ts
```diff
@@ -1,277 +1,179 @@
...
+        query = query.eq("category_id", filters.categoryId)
+      }
+    }    
+
+    if (filterOutDeleted) {
+      query = query.eq("deleted", false)
+    }
+
+    const { data: cvs, error } = await query
+    console.log("cvs:", cvs?.map(cv => cv.category_id))
```
YechiamTK, line 168: Don't forget to delete console logs before pushing.
Also @Aradon267 I'm pretty sure there's either a commit hook or a github actions rule to highlight these logs.

cv_next/server/api/supabaseHelper.ts
```diff
@@ -0,0 +1,71 @@
+import "server-only"
+
+import { createClient, SupabaseClient } from "@supabase/supabase-js"
+import {
+  createBrowserClient,
```
YechiamTK, line 5: Do we have es-lint rule for unused imports?

cv_next/server/api/supabaseHelper.ts
```diff
@@ -0,0 +1,71 @@
...
+      {
+        cookies: {
+          get(name: string) {
+            return cookieStore.get(name)?.value
+          },
+          set(name: string, value: string, options: CookieOptions) {
+            try {
+              cookieStore.set({ name, value, ...options })
+            } catch (error) {
+              // The `set` method was called from a Server Component.
```
YechiamTK, line 53: A. I assume the comment is from their example?
B. Let's actually handle the errors, even a simple log will do for now.

cv_next/components/ui/filters/valueSelect.tsx
```diff
@@ -0,0 +1,45 @@
...
+    text: string
+  }) => {
+    const [isMenuOpen, setIsMenuOpen] = useState(false)
+    const isPlaceHolder = valueId === null
+
+    const changeIsMenuOpen = () => {
+        setIsMenuOpen(!isMenuOpen)
+    }
+
+    return <>
```
YechiamTK: You have multiple conditional rendering flags in an already hard-to-digest tw block of code.
Please separate them as constants above.
e.g. `const menuOpenStyle = isMenuOpen ? "outline-b-0 rounded-b-none" : "";`, then using it later as simply `{menuOpenStyle}`.
Code readability is important, especially with tw classes :)

cv_next/components/ui/filters/valueSelect.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <div className={`flex justify-center items-center relative w-1/6 h-2/6 whitespace-nowrap bg-white outline-gray-40 outline-2 ${isMenuOpen ? "outline-b-0 rounded-b-none" : ""} px-10 py-4 rounded-md cursor-pointer box-border`} onClick={changeIsMenuOpen} >
+            <div className={`${isPlaceHolder ? 'text-gray-400' : 'text-black'}`}>
+                {`${text}: ${isPlaceHolder ? placeHolder : getValueById(valueId)}`}
+            </div>
+            <div className={`absolute z-10 top-full w-full h-fit bg-white block ${isMenuOpen ? "block" : "hidden"} max-h-40 overflow-y-auto border-gray-40 border-2 border-t-0 divide-y-2`}>
+                <div className="text-gray-400 w-full h-10 bg-white hover:bg-slate-200 flex justify-center items-center" onClick={() => onChange(null)}>
```
YechiamTK: BTW in an input component, why aren't we using HTML input?
HTML tags are also very important - divs all the way down is bad practice. I won't comment on every `<article>` or `<aside>` tags, but `<input>` is important enough to note.

cv_next/app/auth/callback/route.ts
```diff
@@ -0,0 +1,37 @@
+import { cookies } from "next/headers"
```
YechiamTK, line 1: `"use server"`?

cv_next/app/auth/callback/route.ts
```diff
@@ -0,0 +1,37 @@
...
+          },
+          remove(name: string, options: CookieOptions) {
+            cookieStore.delete({ name, ...options })
+          },
+        },
+      }
+    )
+    const { error } = await supabase.auth.exchangeCodeForSession(code)
+    if (!error) {
+      return NextResponse.redirect(`${origin}${next}feed`)
```
YechiamTK, line 31: Possible issue: what can `next` be? If it can be anything other than "/", how it relates to "feed"?

cv_next/app/auth/callback/route.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+  }
+
+  // return the user to an error page with instructions
+  return NextResponse.redirect(`${origin}/auth/auth-code-error`)
```
YechiamTK, line 36: Does this route exist?

cv_next/server/api/comments.ts
Aradon267: The entire class still uses Firebase

cv_next/app/feed/components/feed.tsx
Aradon267: We have category-based filters and I think @nevogithub added multi-category filters

cv_next/app/feed/page.tsx
Aradon267: In feed.tsx we do call the action fetchCvsForFeed which uses the getPaginatedCvs from the server

cv_next/app/feed/components/CVItemRSC.tsx
Aradon267: So in this specific case wouldn't it be better to add the changes from there if they're ready?

cv_next/server/api/cvs.ts
Aradon267: Would it be better to change it to the logger that is present in the project? Or just remove it completely? I'll check about the hook

cv_next/server/api/supabaseHelper.ts
Aradon267: Will check

cv_next/server/api/supabaseHelper.ts
Aradon267: I'll add a log for this but technically we do have a middleware

cv_next/app/auth/callback/route.ts
Aradon267: We don't, it redirect's to next's not found
What instructions should this page even contain? It only get's here if the code param wasn't found

cv_next/app/auth/callback/route.ts
YechiamW3: Well, either do that or make sure you can handle other "next" properly. Currently there's a high chance of handling e.g. `next = "/cvs/1"` by redirecting to `/cvs/1feed`. Idk what actual possibilities are for "next", but you should count for them - and if there's only one possibility, then don't use "next" at all (but be sure of it).

cv_next/app/auth/callback/route.ts
YechiamW3: Well, I guess it's an example code right? The point is, don't leave these kind of lines if they don't match any of our actual use case. If we want, we can create this route and handle this case. But if we don't want to handle such case in an informative way, we can simply redirect to error page instead of populating an inexistent route to our codebase.

cv_next/app/feed/components/CVItemRSC.tsx
YechiamW3: Nope. This'll only add more complications to this already enormous branch. Merge this branch and only then open PR to the other branch. There's absolutely nothing wrong with resolving conflicts.

cv_next/app/feed/components/feed.tsx
PeroG28: there suppose to be a search filter too

cv_next/app/auth/callback/route.ts
Aradon267: According to what I see there are 2 options for this:
1. Google decided to add a different redirect path back to our website - that is really unlikely as Google only returns the oAuth code generated for the user.
2. A user decided to add it in the link itself - possible malicious intent. This could be exploitable only if the user also managed to get the code from Google.
I'll make it so instead of `${next}` there's just `/`

cv_next/app/auth/callback/route.ts
Aradon267: This only happens when Google can't authenticate the user(no oAuth code was returned)...
Is that something worth handling?

cv_next/server/api/supabaseHelper.ts
Aradon267: Added precommit for unused imports and vars

cv_next/server/api/cvs.ts
Aradon267: I'll add an eslint commit rule

cv_next/app/feed/components/CVItemRSC.tsx
Aradon267: Ok
Updated the change requested at the start of the thread

cv_next/app/auth/callback/route.ts
YechiamW3: For better maintainability, you can define a `AUTH_DEFAULT_REDIRECT` string in the Definitions file.

cv_next/app/auth/callback/route.ts
YechiamW3: Well sounds like a possible timeout error, etc. Either redirect to some error page we have, or throw to homepage for now.
If we *know* there's a possibility a user will encounter some flow, we'd better handle it. The only possibilities we leave out are those we aren't aware of :)
(this could also be dealt with in a toast-like behavior on the auth page, but it'll require a bit more changes than we actually want right now)

cv_next/app/feed/components/feed.tsx
YechiamW3: What I meant but the "do we have filters" is "can we actually test the thought I had" :)
The point is, there's a possible race for the filter state and the force reload functionality. You'd either want to make a synchronous behavior for the filters state change (absolute hell, don't try and it's an anti-pattern for react), or more appropriately replace it with a `useEffect` and add the filters to the deps array.

cv_next/app/feed/components/filterPanel.tsx
```diff
@@ -0,0 +1,60 @@
+import { useEffect, useState } from "react"
+import { SearchBox } from "../../../components/ui/filters/searchbar"
+import { DropdownInput } from "../../../components/ui/filters/valueSelect"
+import Categories from "@/types/models/categories"
+import { ModeSwitch } from "@/components/ui/filters/modeSwitch"
+
```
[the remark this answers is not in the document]
YechiamW3: Why resolved? Why not delete it? 🤔

cv_next/app/auth/callback/route.ts
Aradon267: Redirecting to the home page, added a todo to redirect to an error page once it is merged

cv_next/app/feed/components/filterPanel.tsx
Aradon267: This will be used in the next version as in this one we decided not to include CV versions.
I've deleted this and the ModeSwitch references but the component itself is still there for the future.

cv_next/components/ui/filters/modeSwitch.tsx
Aradon267: Just note that as I mentioned in a different comment, this isn't used until the next version as we decided to scrap the whole CV versioning idea(for this time)

cv_next/app/feed/components/feed.tsx
Aradon267: I see
Yes we have filters(both for categories and for a description of a CV and it does seem to work fine)

cv_next/app/feed/components/filterPanel.tsx
Aradon267: Updated the onchange code and added memoization for the filtering(which is now also separated)

cv_next/components/layout/navbar.tsx
Aradon267: Don't really know the use of this
As I said, I've deleted it locally and nothing broke or change so deleting these lines

cv_next/server/api/cvs.ts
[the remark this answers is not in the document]
Aradon267: The only things that are exposed are the models, not the strings
Should the names of the columns also be exposed there for each column?

cv_next/app/logout/page.tsx
Aradon267: Moved the signout directly to the popup functionality as this is the referencing place

cv_next/components/layout/navbar.tsx
Aradon267: Cleaned it up a bit(a lot of unused code, hidden divs, etc)
Also moved the popup trigger to another component and created a navbar dir inside layout to aggregate all related components

cv_next/components/ui/filters/modeSwitch.tsx
YechiamW3: OK so two options: a. don't include in the PR (i.e. delete the files from the remote), b. add all of the comments I wrote in these components in a `//TODO` line at the top of the files.
If they'll be merged in their current state, we will simply forget about these comments.

cv_next/app/feed/components/filters/valueSelect.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const menuBorderStyle = isMenuOpen ? "outline-b-0 rounded-b-none" : "";
+  const selectionStyle = isMenuOpen ? "block" : "hidden";
+  return (
+    <>
+      <div
+        className={`outline-gray-40 relative flex h-2/6 w-1/6 items-center justify-center whitespace-nowrap bg-white outline-2 ${menuBorderStyle} box-border cursor-pointer rounded-md px-10 py-4`}
+        onClick={changeIsMenuOpen}
+      >
+        <div className={`${isPlaceHolder ? "text-gray-400" : "text-black"}`}>
```
YechiamW3, line 33: This entire div can be separate (and along the way, be more semantically meaningful in the code context):
```
const textValue = isPlaceHolder ?
    <div className="text-black">
        {`${text}: ${placeHolder}`}
    </div> :
    <div className="text-grey">
        {`${text}: ${getValueById(valueId)}`}
    </div>;

Comment (YechiamW3) [reply cv_next/app/feed/components/filters/valueSelect.tsx:33]: Note that while useful and often times efficient in both code writing time and maintainability, conditional render *inside the return value* can quickly turn into a mess that harms readability. It's a fine line, but you should almost always prefer readability.
Also in this specific case the conditional operator is used twice on the same value, so you are not really utilizing the conditionality in a good maintainable way.

File: cv_next/components/layout/navbar/mode-toggle.tsx [hunk 42]
```diff
@@ -0,0 +1,33 @@
+"use client";
```
YechiamW3, line 1: OK I know you probably didn't write this component as you've only switched dirs here, but pls do the following:
a. change the file name to match the standard component naming (I think this file, along with others, are from some example we took a while back and didn't handle correctly; pls see other component files that are with snake-case instead of camelCase).
b. don't `import * as React`, import the specifics needed.
c. we have a `isBrowser` util in the lib folder, use it instead of the `isClient` state.

cv_next/types/database.types.ts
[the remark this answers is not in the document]
Aradon267: Only saw this now, updating to server only

cv_next/components/layout/navbar/mode-toggle.tsx
Aradon267: Changed all notes
Also check for other import * and fixed these

cv_next/app/feed/components/CVItemRSC.tsx
Aradon267: Managed to work fine with plaiceholder, added that

cv_next/components/layout/popup.tsx
Aradon267: Switch to the HTML dialog and added the click outside functionality

cv_next/components/layout/navbar/popupToggle.tsx
```diff
@@ -0,0 +1,26 @@
+"use client";
+import settignsIcon from "@/public/images/settigns.png";
+import Image from "next/image";
+import { useState } from "react";
+import Popup from "./popup";
+
+export function PopupToggle() {
+  const [isSettignsOpen, setIsSettignsOpen] = useState(false);
```
YechiamW3, line 8: OK sorry but I see too many "settingns", pls search-and-replace it :)

cv_next/app/feed/components/CVItemRSC.tsx
```diff
@@ -1,18 +1,35 @@
...
+const getBlur = cache(async (imageURL: string) => {
+  return await getBlurredCv(imageURL);
+});
 interface CVCardProps {
-  cv: ClientCvModel
+  cv: CvModel;
 }
 
 export default function CVItemRSC({ cv }: CVCardProps) {
+  const [base64Data, setBase64Data] = useState(
```
YechiamW3, line 17: By adding this you've forced it to be client component :)
you can simply do something like `getBlur(imageUrl) ?? '...';` instead of introducing state.
By the way, because this component is a server component, the browser should already receive it with the data loaded - that's the point of loading the data on the server. It's not the browser sending the request, it's the server sending it, before sending the component (html/js code) to the browser. So you should be fine with:
```Suggestion
const base64Data = await getBlur(imageUrl);
```

Just make sure that `getBlur` returns string even on failure.

Aradon267: Tried doing it the way you mentioned(removed all references to useState/Effect, added await to getBlur and async near the export of the export, etc) but it said that async/await is only supported in server components and every time new items loaded, the entire thing disappeared. Is that the way you meant?
When I tried adding use server it crashed because it couldn't handle the Props.

cv_next/components/ui/filters/valueSelect.tsx
Aradon267: Are you sure input is the correct thing here? This is supposed to be for the categories, not the user input(that's in searchBar which does use input)

cv_next/app/feed/components/CVItemRSC.tsx
YechiamW3: Show me a) how the file looks now and b) how the component that calls it looks

cv_next/app/feed/components/filterPanel.tsx
```diff
@@ -0,0 +1,59 @@
...
+}) => {
+  const [searchValue, setSearchValue] = useState(defaultFilters.searchValue);
+  const [categoryId, setCategoryId] = useState(defaultFilters.categoryId);
+
+  useEffect(() => {
+    onChange({
+      categoryId: categoryId,
+      searchValue: searchValue,
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
YechiamW3, line 30: There's no actual reason to ignore the exhaustive deps warning here. It's not really important, but ignoring exhaustive deps should be only in extreme cases, not a common practice. You can easily ignore something rather important you don't really know of its importance if you're used to ignoring.

Aradon267: I ignored it due to the spam of reviewbot over it(as it is important) but in this case the `onchage` function that it talks about doesn't change

cv_next/app/feed/components/CVItemRSC.tsx
Aradon267: The RSC, with use server at the top
```ts
import { cache } from "react";
import { getBlurredCv } from "@/app/actions/cvs/getBlurCv";

const getBlur = cache(async (imageURL: string) => {
  return await getBlurredCv(imageURL);
});

interface CVCardProps {
  cv: CvModel;
}

export default async function CVItemRSC({ cv }: CVCardProps) {
  const categoryLink = `/feed?category=${Categories.category[
    cv.category_id
  ].toLowerCase()}`;
  const imageUrl = generateImageUrl(cv.document_link);
  const formattedDate = new Date(cv.created_at).toLocaleDateString("en-US");
  const base64Data =
    (await getBlur(imageUrl)) ??
    "data:iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mPs7p5fDwAFlAI2LB7hbAAAAABJRU5ErkJggg==";

  return (...)
```
The ... at the return is just the image with the data the also needs to be with the image(category, uid, etc)

What's calling in is the feed that is pretty long but the relevant part is
```ts
cvs.map((cv) => (
              <CVItem key={cv.id} cv={cv}>
                <CVItemRSC cv={cv} />
              </CVItem>
            )
```

I can add the files if you'd like

cv_next/app/feed/components/feed.tsx
Aradon267: Tagging here just t see if a follow up is needed here @YechiamW3

cv_next/app/feed/page.tsx
Aradon267: Tagging here just to see if a follow-up is needed @YechiamW3

cv_next/server/api/cvs.ts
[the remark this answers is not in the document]
Aradon267: Tagging here just to see if a follow-up is needed @YechiamW3

cv_next/app/feed/components/CVItemRSC.tsx
YechiamW3: The reason I wanted to see the file you import it in, is because client and server interleaving is a bit complicated. Make sure you follow this pattern: https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#supported-pattern-passing-server-components-to-client-components-as-props.
I recommend you read the whole page and make sure you understand it.

cv_next/app/feed/components/filterPanel.tsx
YechiamW3: The reviewbot's messages are not "spam", they're here to help us :)
I know that *currently* there's absolutely no actual need to put the on change function in the deps array, but because it's good practice to put everything *subject to change* in the deps array (because you never know what the code would look like in the future), we do want to follow it. You can look it up, but generally the consensus is always listen to the exhaustive-deps warn.

cv_next/app/feed/components/feed.tsx
YechiamW3: I think that we can leave it, but there's still a racing risk here.
The best way to solve it is to do as I've suggested.

cv_next/server/api/cvs.ts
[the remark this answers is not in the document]
YechiamW3: Sorry I probably missed it. In the most general sense - strings should be constants, unless used only singularly (i.e. in a text block display). I'm not with a computer right now so I can't look, but I believe "cvs" is exposed in some manner under the model..

cv_next/app/feed/components/filterPanel.tsx
Aradon267: I see. I tried putting that the onchange there but it kept reloading everything, that’s why I commented that, but I’ll make sure I delete it. Any idea about the reload thing?

cv_next/server/api/cvs.ts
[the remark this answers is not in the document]
Aradon267: I’ll try to see if it’s defined somewhere but if it’s not, should I create some constants in the global for the entire thing?

cv_next/app/feed/components/filterPanel.tsx
YechiamW3: Ah, great, it means you probably need to wrap it using useCallback because it does, in fact, changes on every render :)

cv_next/server/api/cvs.ts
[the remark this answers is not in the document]
YechiamW3: Well, Avner preferred the global but I personally prefer specific definition files. That being said, I'll let you decide what you prefer :)
But do be consistent.

cv_next/app/feed/components/filterPanel.tsx
Aradon267: Yeah lol
A bit after writing that I check some stuff and saw that’s the way to go

cv_next/app/feed/components/feed.tsx
Aradon267: Added your suggestion

cv_next/app/feed/components/CVItemRSC.tsx
Aradon267: Read it several times, couldn't understand how to send the props to CVItemRSC from the parent(or skip it and pass only to the children)

PR #65, opened 2024-06-08
Search text

nevogithub:


nevogithub:
It has some search textss and users

.vscode/launch.json
YechiamW3, line 1: shouldn't this folder be in gitignore?

cv_next/.env
YechiamW3, line 1: Pretty sure this shouldn't be pushed. @Aradon267?

cv_next/server/api/cvs.ts
```diff
@@ -164,7 +169,10 @@ export async function getPaginatedCvs(
     if (filterOutDeleted) {
       query = query.eq("deleted", false);
     }
-
+    if(checkNames){
+      ids = await queryUserId;
```
YechiamW3, line 173: Is this working?
I think it should be `{ data: ids }`...

YechiamW3: Oh I see you simply don't extract data from the result; it's preferable to do so. See the other functions. Lets you early-exit if you encounter an error.

.vscode/launch.json
Aradon267: This should

cv_next/.env
Aradon267: Same for that yes

cv_next/server/api/cvs.ts
```diff
@@ -4,7 +4,7 @@ import MyLogger from "@/server/base/logger";
 import Categories from "@/types/models/categories";
 import Definitions from "../../lib/definitions";
 import SupabaseHelper from "./supabaseHelper";
-import { PostgrestError } from "@supabase/supabase-js";
+import { PostgrestError, } from "@supabase/supabase-js";
```
PeroG28, line 7: delete, this shold deleted by the formater (if we have one)

cv_next/server/api/cvs.ts
```diff
@@ -4,7 +4,7 @@ import MyLogger from "@/server/base/logger";
...
 import { filterValues } from "@/app/feed/components/filterPanel";
 
 export const revalidate = Definitions.CVS_REVALIDATE_TIME_IN_SECONDS;
```
PeroG28, line 10: REVALID_DATE, global  const needs to be capitel with _ between words

cv_next/server/api/cvs.ts
```diff
@@ -147,13 +147,18 @@ export async function getPaginatedCvs(
     const supabase = SupabaseHelper.getSupabaseInstance();
     let query = supabase
```
PeroG28, line 148: why this is not one query with join?

cv_next/server/api/cvs.ts
```diff
@@ -147,13 +147,18 @@ export async function getPaginatedCvs(
...
+      .select("*");
+    let queryUserId = supabase
+      .from("profiles")
+      .select("*");
+    let checkNames = false;
+    let ids = null;
     MyLogger.logDebug("filters", filters);
     if (filters) {
       if (filters.searchValue) {
         query = query.textSearch("description", filters.searchValue);
```
PeroG28, line 159: try to combine queries, maybe there is a way to text search 2 fileds

cv_next/server/api/cvs.ts
Aradon267: This isn't used so can be deleted
and revalidate is one word

cv_next/server/api/cvs.ts
nevogithub: Yeah I'll fix it to match the error handling pattern

cv_next/server/api/cvs.ts
nevogithub: IMHO it should be a seperate query as we;re querying different data. I've been working on a function to query IDs from fullnames seperately

Aradon267:
@nevogithub is everything alright with this? sent you on WhatsApp the stable filter version as we discussed

Aradon267:
As discussed in the last weekly, I'll handle things here

PR #73, opened 2024-06-28
72 fix session management with ssr

Aradon267:


Aradon267:
@YechiamW3 This is ready for review, but not exactly ready yet:
This branch fixes the problem with changing the cookies in the server(the one that Nevo brought up)
It does so by changing the entire login process to be on the server side(which meant making SupabaseHelper not a singleton).
The reason this isn't fully ready yet is the middleware.  Until now, the middleware used supabase's `getUser` to get the current user from the server and see the user's level using a select. This started causing rate limits so I moved the level check to only happen when a page loads and removed it from the middleware(the middleware had no interaction with the app at that point) which caused the cookies issue. Apparently, what caused the rate limit was 'getUser' which was triggered for every server action under feed(could be 100+ in a matter of 1 minute) so I switched to 'getSession' which fixed everything in this branch.
The problem is that using `getSession` is insecure and could be exploited easily according to a [this](https://github.com/orgs/supabase/discussions/23224). The link gave 2 ways of mitigation:
1. Use 'getSession' and validate the `access_token` using a 3rd party library(and the JWT secret which is not that great either) - this was still not that secure because I could use my access_token with your ID and I can get your data so I added a check on the userId provided from the session and on the userId from the decoded token which caused `Using the user object as returned from supabase.auth.getSession() or from some supabase.auth.onAuthStateChange() events could be insecure! This value comes directly from the storage medium (usually cookies on the server) and many not be authentic. Use supabase.auth.getUser() instead which authenticates the data by contacting the Supabase Auth server.` to get printed for every server action because every server action would go through the middleware and get the session checked out. This did however handle the rate limit. We also can decide that the validation of the token on its own is enough to see if the user is logged in and do the other auth checks in a different place.
2. Use `getUser` which caused a rate limit although someone on supabase's discord told me that it's cached as it's just a fetch request to the server so I'll try to understand more about that.

If you have any suggestion on this please tell

cv_next/app/actions/user/fetchUserInfo.ts
```diff
@@ -49,3 +58,18 @@ export const getUserFromId = async (
...
+    await SupabaseHelper.getSupabaseInstance().auth.signInWithOAuth({
+      provider,
+      options: {
+        redirectTo:
+          process.env.NEXT_PUBLIC_BASE_URL + Definitions.AUTH_CALLBACK_REDIRECT,
+      },
+    });
+  if (error) logger.error(error, "Error signin");
+  if (data.url) {
+    redirect(data.url); // use the redirect API for your server framework
```
YechiamW3, line 73: delete comment :)

cv_next/app/feed/components/CVItem.tsx
```diff
@@ -50,37 +50,41 @@ export default function CVItem({ cv }: CVCardProps) {
   );
 
   useEffect(() => {
+    let isMounted = true;
```
YechiamW3, line 53: Are you certain this locally-scoped var is giving you what you need? It seems suspicious to me. I think it'd be preferrable if you'd turn this into a ref variable.

YechiamW3: Also if all you want is for this effect to run only during mount, just empty the deps array

YechiamW3:
> @YechiamW3 This is ready for review, but not exactly ready yet: This branch fixes the problem with changing the cookies in the server(the one that Nevo brought up) It does so by changing the entire login process to be on the server side(which meant making SupabaseHelper not a singleton). The reason this isn't fully ready yet is the middleware. Until now, the middleware used supabase's `getUser` to get the current user from the server and see the user's level using a select. This started causing rate limits so I moved the level check to only happen when a page loads and removed it from the middleware(the middleware had no interaction with the app at that point) which caused the cookies issue. Apparently, what caused the rate limit was 'getUser' which was triggered for every server action under feed(could be 100+ in a matter of 1 minute) so I switched to 'getSession' which fixed everything in this branch. The problem is that using `getSession` is insecure and could be exploited easily according to a [this](https://github.com/orgs/supabase/discussions/23224). The link gave 2 ways of mitigation:
> 
> 1. Use 'getSession' and validate the `access_token` using a 3rd party library(and the JWT secret which is not that great either) - this was still not that secure because I could use my access_token with your ID and I can get your data so I added a check on the userId provided from the session and on the userId from the decoded token which caused `Using the user object as returned from supabase.auth.getSession() or from some supabase.auth.onAuthStateChange() events could be insecure! This value comes directly from the storage medium (usually cookies on the server) and many not be authentic. Use supabase.auth.getUser() instead which authenticates the data by contacting the Supabase Auth server.` to get printed for every server action because every server action would go through the middleware and get the session checked out. This did however handle the rate limit. We also can decide that the validation of the token on its own is enough to see if the user is logged in and do the other auth checks in a different place.
> 2. Use `getUser` which caused a rate limit although someone on supabase's discord told me that it's cached as it's just a fetch request to the server so I'll try to understand more about that.
> 
> If you have any suggestion on this please tell
> 
> After talking to some people in supabase, is it possible that we have a render loop that makes it so each time we get new cvs to the cvs array we re-render the entire thing?

I believe multiple re-renders are probably the root issue. You should try to investigate this first. Simplest way is to add a useEffect with a log and see how many times the log is printed. e.g.:
```
const useEffect(() => {
    console.log("Component Rendered");
},[]);
```

cv_next/app/feed/components/CVItem.tsx
Aradon267: I want it to run only once but I want to use the interval to revalidate the image every 2 minutes.
Reason I added this was that when I signed out or move to a different page, it would tell me that it can't find a server action with that ID

Aradon267:
It seems like each cv is being rendered twice and that feed is also being rendered twice

cv_next/app/feed/components/CVItem.tsx
YechiamW3: I see, so the cleanup function should only run on unmount (or in legacy react terms, willUnmount)? Note that this is not the functionality you've implemented.
The cleanup function is called on every re-render. See https://react.dev/reference/react/useEffect#useeffect - "After every re-render with changed dependencies, React will first run the cleanup function (if you provided it) with the old values...".
What you want is to separate this into multiple effects, let's say first for the current version you have and a second for unmounting (only the return function with no deps).
BTW a better method instead of the `isMounted` check, imo, will be to take the interval ID as a ref and clear it on the unmounting effect (https://developer.mozilla.org/en-US/docs/Web/API/clearInterval).

YechiamW3:
> It seems like each cv is being rendered twice and that feed is also being rendered twice... The people from Supabase Discord changed their minds and believed that the root cause of the rate limit problem might even be the fact that we use server actions for data fetching:
> 
> ```
> However, if you're running calling all these server actions client side to fetch data then that seems more like the issue.
> Sounds like this data should either be;
> a) fetched client side (no server action involved).
> b) fetched on the server (in a server component) and passed down as props.
> 
> I don't think server actions are for data retrieval, but rather for data mutations.
> ```
> 
> ```
> Seems like you're using server actions to fetch data client side, this is not the intended purpose.
> 
> Server actions are for data mutations, server side or client side
> Server components are for data fetching
> Anything else you can do as you would with any other react app before these were introduced.
> 
> So if you're calling a server action client side for each cell etc etc, you're hitting the middleware because the server action posts directly to the current route.
> ```

Well, you can ask them how a data state should be maintained using both server and client logic. The point is that Next is more of a mess than I initially thought.
The "suggested" solution, as we've talked about before, is pre-fetching on the server an initial payload (e.g. first page) using server action, then load additional data using client fetching. Server actions are not that different from route handlers for this. I truly don't know what they're getting at. Please explain to them that your data fetching is dynamic (you can share the specifics - continuous information feed that triggers additional fetching based on window location).
If they mean to use "regular" api calls, it seems wrong to me. The point is to let the server handle the db connection and not connect to sb directly from the client.
Anywho, you can try two things and see if it helps:
1. Replace server action with Route Handler
2. Replace server action with client-side api call directly to supabase (i.e. supabase-client)
Although, even if any of the two helps, this shouldn't *be* a problem, it's a supabase problem.
Literally any infinite-scroll guide suggests this method (e.g. https://blog.logrocket.com/implementing-infinite-scroll-next-js-server-actions/#setting-up-server-action).

I actually see that many of the articles that talk about infinite scroll *with supabase* handle the data fetching with client-side supabase calls. This is triggering. The issue is 100% with supabase. Let me know what they're saying, but if we'll have no choice than switch to client supabase calls.

Aradon267:
I asked them about so direction on the topic, mainly if they know a way to do a infinite scroll like that:

```
I'm pretty sure it's a scenario where just fetching client side is still the best option. Or I think you could also do them using server component and a suspense boundary. That might actually be the 'new correct way' if you want to double down on RSC's.

 So you probably have a scroll  container as a client component because you listen to scroll events etc, then you pass the content component as a server component as a prop to the client component, and then when the user scrolls you mount the RSC with a suspense bouncary around it. But it's all theoretical I haven't tried it with infinite scrolling.
```

Also I saw this link to a guide of sb about it

https://supabase.com/blog/infinite-scroll-with-nextjs-framer-motion

Aradon267:
Technically the recommended way to use supabase is to always recreate the server client with the relevant cookies for each request(as seen in the code I updated) while in client the client is a singleton and either way there always technically were 2 connections to supabase

YechiamW3:
> Technically the recommended way to use supabase is to always recreate the server client with the relevant cookies for each request(as seen in the code I updated) while in client the client is a singleton and either way there always technically were 2 connections to supabase Moving everything to server components is impossible due to the infinite scroll unless we somehow manage to send the data down as props from the server(which I think isn't possible) which might not trigger the middleware. Maybe really for this moving even just the fetching to client is the way to go... I'll have to think about the plaiceholder thing(it must be in a server component) but I'll try to manage

Wait, where are we using the client supabase util (i.e. `createClientComponent`)?
I don't see any usage of it. As I'd expect, as a db connection should *generally* be maintained on the server.
But alas, if sp doesn't allow us to have proper code, we'll give in and put everything on the client-side access. We would want significant security checks, but I trust that the client access is secure enough.
Anyway, regarding other server-side concerns: I don't see how it's related to sp connection. Unless what you mean with the middleware is that we can't do *any* server request without triggering rate limits? If so, then what they're saying is we should abandon *any* server request. It doesn't sound right.

Aradon267:
> 
> Wait, where are we using the client supabase util (i.e. `createClientComponent`)? I don't see any usage of it. As I'd expect, as a db connection should _generally_ be maintained on the server. But alas, if sp doesn't allow us to have proper code, we'll give in and put everything on the client-side access. We would want significant security checks, but I trust that the client access is secure enough. Anyway, regarding other server-side concerns: I don't see how it's related to sp connection. Unless what you mean with the middleware is that we can't do _any_ server request without triggering rate limits? If so, then what they're saying is we should abandon _any_ server request. It doesn't sound right.

Don't think we're using it anywhere at the moment but we will start soon hopefully.
About the server connection, I think we just need to do better when it comes to how we call the server actions maybe using regular fetch calls i.e. HTTP://localhost ... /api/do/something and we can even hide it from the middleware by excluding this specific path

YechiamW3:
> > Wait, where are we using the client supabase util (i.e. `createClientComponent`)? I don't see any usage of it. As I'd expect, as a db connection should _generally_ be maintained on the server. But alas, if sp doesn't allow us to have proper code, we'll give in and put everything on the client-side access. We would want significant security checks, but I trust that the client access is secure enough. Anyway, regarding other server-side concerns: I don't see how it's related to sp connection. Unless what you mean with the middleware is that we can't do _any_ server request without triggering rate limits? If so, then what they're saying is we should abandon _any_ server request. It doesn't sound right.
> 
> Don't think we're using it anywhere at the moment but we will start soon hopefully. About the server connection, I think we just need to do better when it comes to how we call the server actions maybe using regular fetch calls i.e. [HTTP://localhost](HTTP://localhost) ... /api/do/something and we can even hide it from the middleware by excluding this specific path

Why do we need the client sp component? (i mean, aside from this issue)
In sp world, is there a difference between data fetching from the browser or from the server, in terms of e.g. security, db connections etc? If not, we can exclude that data fetching paths from the middleware (which I hope is possible for server actions), or we can switch data fetching to the browser. But from what I understand, keeping the middleware in place for mutation actions is justified, and hopefully won't give us rate limits (we can minimize requests from the browser with debounce for example).
For plaiceholder, see this suggestion: https://github.com/vercel/next.js/discussions/26168#discussioncomment-886353
We've talked about something similar before. I believe it's either that, or pre-fetching a big number of cvs beforehand. Which, if we're being frank, would take a long time to scale: I don't think we expect more than 100 cvs in the first few months; so pre-fetching initial load of 100, UX-ing infinite scroll for these, and then load another 100 (can even load in background before reaching the 100) - would be pretty good overall.
It also seems like a severely low percentage of visits that would *actually* scroll for more than 100..
Anyway, just a thought, but it might be a good direction.
Either way I believe that pre-rendering the placeholders and statically saving them will be nice.

Aradon267:
> Why do we need the client sp component? (i mean, aside from this issue)
> 
> In sp world, is there a difference between data fetching from the browser or from the server, in terms of e.g. security, db connections etc? If not, we can exclude that data fetching paths from the middleware (which I hope is possible for server actions), or we can switch data fetching to the browser. But from what I understand, keeping the middleware in place for mutation actions is justified, and hopefully won't give us rate limits (we can minimize requests from the browser with debounce for example).
> 
> For plaiceholder, see this suggestion: https://github.com/vercel/next.js/discussions/26168#discussioncomment-886353
> 
> We've talked about something similar before. I believe it's either that, or pre-fetching a big number of cvs beforehand. Which, if we're being frank, would take a long time to scale: I don't think we expect more than 100 cvs in the first few months; so pre-fetching initial load of 100, UX-ing infinite scroll for these, and then load another 100 (can even load in background before reaching the 100) - would be pretty good overall.
> 
> It also seems like a severely low percentage of visits that would *actually* scroll for more than 100..
> 
> Anyway, just a thought, but it might be a good direction.
> 
> Either way I believe that pre-rendering the placeholders and statically saving them will be nice. 

Client is used mainly in checks about who the user is(if I'm the user that commented, if this is my profile, etc - which thinking about it now, this is not that secure either)
I think that there is some difference because of the cookies that are used(client side could be altered)...
Excluding functions in the middleware is impossible as far as I saw. For mutation everything is fine.
Instead of picking the maybe unsafe option and switch everything to client, should I try and see if changing things to api routes and excluding them in the middleware help?
I'll see what I can do about the pre fetching

YechiamW3:
> > Why do we need the client sp component? (i mean, aside from this issue)
> > In sp world, is there a difference between data fetching from the browser or from the server, in terms of e.g. security, db connections etc? If not, we can exclude that data fetching paths from the middleware (which I hope is possible for server actions), or we can switch data fetching to the browser. But from what I understand, keeping the middleware in place for mutation actions is justified, and hopefully won't give us rate limits (we can minimize requests from the browser with debounce for example).
> > For plaiceholder, see this suggestion: [vercel/next.js#26168 (reply in thread)](https://github.com/vercel/next.js/discussions/26168#discussioncomment-886353)
> > We've talked about something similar before. I believe it's either that, or pre-fetching a big number of cvs beforehand. Which, if we're being frank, would take a long time to scale: I don't think we expect more than 100 cvs in the first few months; so pre-fetching initial load of 100, UX-ing infinite scroll for these, and then load another 100 (can even load in background before reaching the 100) - would be pretty good overall.
> > It also seems like a severely low percentage of visits that would _actually_ scroll for more than 100..
> > Anyway, just a thought, but it might be a good direction.
> > Either way I believe that pre-rendering the placeholders and statically saving them will be nice.
> 
> Client is used mainly in checks about who the user is(if I'm the user that commented, if this is my profile, etc - which thinking about it now, this is not that secure either) I think that there is some difference because of the cookies that are used(client side could be altered)... Excluding functions in the middleware is impossible as far as I saw. For mutation everything is fine. Instead of picking the maybe unsafe option and switch everything to client, should I try and see if changing things to api routes and excluding them in the middleware help? I'll see what I can do about the pre fetching

If api routes can be excluded - yeah let's try that.

Aradon267:
I've uploaded a version that uses API paths only for the fetch and things that are related to the preview which are ignored in the middleware, lmk what you think
I also did some tests and I think that the main cause of the rate limit of the getUser was caused due to the fact that the filters that are text-based, immediately search for results(when I type "hello" it'll search "h" "he" "hel" ...)

cv_next/app/feed/components/CVItem.tsx
Aradon267: I think I understood what you meant, I uploaded a different version

cv_next/app/actions/cvs/fetchCvs/route.ts
```diff
@@ -0,0 +1,22 @@
+"use server";
```
YechiamW3, line 1: We should keep things separates; `actions` folder is for server actions, api routes should be under a different folder - e.g. in a different `api` folder

cv_next/app/actions/cvs/fetchCvs/route.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { filterValues } from "@/app/feed/components/filterPanel";
```
YechiamW3, line 3: If you're using it outside the filter panel component, please move it to a more general location

cv_next/app/feed/components/CVItem.tsx
```diff
@@ -27,62 +22,107 @@ export default function CVItem({ cv }: CVCardProps) {
 
   const getBlur = useMemo(
     () => async (imageURL: string) => {
-      return await getBlurredCv(imageURL);
+      const response = await fetch("/actions/cvs/fetchPreviews", {
```
YechiamW3, line 25: Relevant to all the fetching functions here: a) please replace the hardcode uris with constants; preferably you can create a route uri generator in the utils. b) please switch to hooks, or some form of encapsulation for these methods as they unnecessarily clutter up this component.

cv_next/middleware.ts
```diff
@@ -1,13 +1,16 @@
...
 import { NextResponse } from "next/server";
 import type { NextRequest } from "next/server";
-import { jwtVerify } from "jose";
 
 export async function middleware(request: NextRequest) {
   let response = NextResponse.next({
     request,
   });
 
+  if (request.nextUrl.pathname.startsWith("/actions/cvs/fetch")) {
```
YechiamW3, line 10: Here again, don't forget to remove hardcoded value

Aradon267: Resolving this as I updated according to the last comment but want to hear your opinion on this - combining this and upper comment

cv_next/app/api/cvs/fetchCvs/route.ts
```diff
@@ -0,0 +1,21 @@
+"use server";
+
+import { filterValues } from "@/types/models/filters";
+import { getPaginatedCvs } from "@/server/api/cvs";
+import { NextRequest, NextResponse } from "next/server";
+
+export async function POST(req: NextRequest) {
+  try {
+    const data = await req.json();
+    const { nextPage, filters }: { nextPage?: number; filters: filterValues } =
```
YechiamW3, line 10: Do we have any user validation on these routes? We want either user validation, or for these safe routes at the very least a rate limit (either here or, if possible, on the platform itself).

cv_next/app/api/users/fetchUserData/route.ts
```diff
@@ -0,0 +1,36 @@
+"use server";
+
+import SupabaseHelper from "@/server/api/supabaseHelper";
+import logger from "@/server/base/logger";
+import { Tables, ProfileKeys } from "@/lib/supabase-definitions";
+import { NextRequest, NextResponse } from "next/server";
+
+export async function POST(req: NextRequest) {
```
YechiamW3, line 8: Here for example we must have some sort of user validation

cv_next/app/feed/components/CVItem.tsx
```diff
@@ -65,20 +92,33 @@ export default function CVItem({ cv }: CVCardProps) {
...
+    cv.user_id,
+    getBlur,
+    getCachedUserName,
+    getURL,
+    revalidatePreview,
+  ]);
+
+  useEffect(
+    () => () => {
+      if (interval != null) {
```
YechiamW3, line 116: I believe if you'll do the check on `interval.current` ts won't require you to cast as Timeout on the next line

YechiamW3:
> I've uploaded a version that uses API paths only for the fetch and things that are related to the preview which are ignored in the middleware, lmk what you think I also did some tests and I think that the main cause of the rate limit of the getUser was caused due to the fact that the filters that are text-based, immediately search for results(when I type "hello" it'll search "h" "he" "help" ...)
> 
> By the way, if we now use fetch for fetching, should we maybe switch to SWR too like with the comments or should we keep the regular fetch?

On the second point - this is very easily solvable with debouncing. If nextjs doesn't have one, either create one or import from lodash.

And yes, if we opt to SWR we should opt in all the way, that way the caching handling will be unanimous.

cv_next/hooks/useAPIFetch.ts
```diff
@@ -0,0 +1,33 @@
+import { useCallback } from "react";
+
+export const CVS_API_BASE = "/api/cvs";
```
YechiamW3, line 3: These should probably sit in definitions rather than the specific api files; something like API_DEFINITIONS.

cv_next/app/api/users/fetchUserData/route.ts
Aradon267: This apparently is a bit problomatic
If we want to have RLS on the profiles that checks if someone is active, how is it that the active status is in the profiles table?

YechiamW3: I see what you mean; but we have 0auth authentication right? I forgot, what's our auth flow? This here should be available after a sign-in, so essentially we should have some 0auth creds we can check against (jwt or something similar).
This is what I mean by the user validation.

Aradon267: I see what you mean, I can add a simple test just to see that we have a JWT but checking the JWT's content will lead to the error that the API routes solved...

YechiamW3: Wait, walk me through the flow. We get login through oauth, and what claims do we get from there? What info do we have that we later use in the "user-handling" process? Username? UserId?
I have an idea that is not very smart but can at least help us bypass the stupid supabase bullshit. We can create a simple list file on the deployment server (not in supabase) that dups the relation we need to match the user and their access level. It'd be referable if we can use uuids when checking against this list. It should be updated either in a cron job, or straight via the user-handling actions.
It's a silly data duplication that might have issues, but we have to handle this somehow. We can't leave these routes unguarded..

Aradon267: 1. Login using oAuth of google
2. Getting a code from google and exchanging it to get an active session from supabase
3. In the cookies we see several things:
- Access Token which is encrypted with a key we have in supabase. The decrypted version has all of the other things that we can find in the cookie.
- User-related stuff like id, email, name, photo, etc.
4. For each supabase request, we send the cookie which includes the id. Using supabase's RLS rules we can get the id, see the "user_type" which specifies the permissions we gave the user that made the request as we don't want anyone that logs in to simply have everything.
5. If the user is at least "active", it can have read access to everything but can't edit things like cvs that aren't his, or comments, etc.
The full list is in another issue if you want to see it.
What I suggest is this:
1. Create a separate table for relations between userid and the perms which will be update with a trigger similar to how we create the user's entry in our current table.
2. Set the RLS of that table to only allow select to everyone.
3. Each RLS of our tables will see if the user is in that table as active or above

YechiamW3: And the problem with supabase that made us go to routes? In which step is it? 4?
Wait, the RLS rules are applied on all user requests? i.e. if a request is sent to this route (and all routes), perhaps we won't know if the user is allowed to perform this action, but supabase will not allow it?
If that's the case, perhaps we can simply rely on the supabase RLS rules, and only correctly handle the errors received from it.
(and if that's what you meant to do in the first place, sorry I didn't understand it)

Also I've checked our RLS - as I understand it, it draws from "whitelisted" and "admins" views - "whitelisted" currently is all the users and not active and above. Just an FYI, you probably know.


Oh, and preferably catching unauthorized requests as early as possible is what should guide us, but the other option is to basically provide free access to all user accounts (even if it's just ids and access level), which I'm not sure is better.

cv_next/app/actions/cvs/fetchCvs.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { getCvById } from "@/server/api/cvs";
+
+export const getCvsFromComments = async (
+  comments: CommentModel[]
+): Promise<CvModel[]> => {
+  const uniqueDocuments: Set<CvModel> = new Set(); // Use a Set to store unique documents
+
+  // Iterate through the comments
+  for (const comment of comments) {
```
YechiamW3, line 11: You might be adding only unique documents, but you're making a call for a cv on each comment (at least each parent comment - which is probably the "busiest" level).
Try instead to filter the unique comment ids beforehand (transform cv ids list from comments list -> use set on that).


Also - if this is related to a different PR and is here only because of a merge - please don't merge branches until we approve a PR, I don't know directly how to associate each change and we're doing unnecessary double work. Let's focus on resolving the issues specific to each PR, and then when the PR is approved merge and resolve conflicts (you can either leave a message that you only merged a branch and will merge the PR later, or simply merge the PR right after).

Aradon267: This is from profiles that was already merged
Will update according to what you said

cv_next/app/api/users/fetchUserData/route.ts
Aradon267: We switched to routes to api using server actions so we could ignore them in the middleware so technically yes 4
RLS works on every access to the DB - in the backend we don't check perms because supabase does it for us with RLS.
whitelisted is active and above we just dont have any inactive users
The problem with catching it as soon as possible is that to see unauthorized requests we need to see who is unauthorized which means getting the data of the profiles i.e. id and type so we need to allow public select access to this
The main problem is that I couldn't get the correct RLS to allow me to get the profile data on profiles that weren't me without full access
If full access:
Got everything
With checking if I'm whitelisted:
Got only myself

YechiamW3: Wait, why do you need more than yourself? Also, can you confirm the behavior I talked about - that currently all of our routes will get blocked in supabase level via the RLS if the user is not active and above? Which means that (albeit not the best solution, as we've talked about early-as-possible) we can rely on the RLS and simply catch the errors here, which means we don't "need" another level of security on the route-level (as I initially thought).

Actually on second thought, perhaps making the profile uuid-type data public might not be _that_ risky (uuids are already essentially public, and user type doesn't need to be secure). I'll admit I'm not that nifty on the security side so asking Alik might be good. Perhaps I'll ask Niki as well. What do you think?

Aradon267: We need others to see their uid to see data about them for the profile/full name of commenter or uploader
Yes as of right now each path will get blocked at supabase level according to user type(active and above). The only exception is as mentioned the profiles which allows select to everyone. Other than the profiles we can trust rls.
Asking them can be a good idea.
Something I just remembered is that profiles also contain if that person is looking for a job and on which category which might be a bit unsafe…

YechiamW3: On the first point - you mean on the normal flow when viewing e.g. a cv page, user with "active" type (i.e. non-admin) will get rejected for requests on profiles for comments? Why would that happen? The profiles table is defined for whitelisted in the RLS, no?
And I'll try asking them we'll see what they have to say.

Aradon267: Currently we have RLS public true for selecting profiles
If I change it I can only see myself(I won't be able to fetch other usernames or ids and stuff like that
Whitelisted is only for the id and type
Notice that there's a policy for public which is just true

Aradon267: Apparently you were correct about the whitelist thing, updated it

cv_next/hooks/useAPIFetch.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  if (!response.ok) {
+    throw new Error(`Failed to fetch data from ${url}`);
+  }
+
+  return response.json();
+};
+
+export const useApiFetch = () => {
```
Aradon267, line 22: Set timeout for API requests

Aradon267:
I think that until we find a way to reduce the number of re-renders we don't really have another way than the API and the perms tables

PR #81, opened 2024-07-20
Upload cv

Aradon267:


cv_next/app/feed/components/button.tsx
```diff
@@ -0,0 +1,23 @@
...
+}: {
+  bottomFlat?: boolean
+  text: ReactNode;
+  onClick: () => void;
+  isDisabled?: boolean
+}) => {
+  const menuBorderStyle = bottomFlat ? "outline-b-0 rounded-b-none" : "";
+  return (
+    <div
+          className={`outline-gray-40 text-black flex h-full w-full items-center justify-center whitespace-nowrap bg-white ${isDisabled ? 'opacity-25' : 'cursor-pointer hover:bg-slate-200'} outline-2 ${menuBorderStyle} box-border rounded-md px-10 py-4`}
```
YechiamW3, line 17: It's better to wrap a `<button>` element, if you're creating a button component.. (in this method you're missing all the button attributes)

cv_next/app/feed/components/inputbar.tsx
```diff
@@ -0,0 +1,48 @@
+export const InputBox = ({
```
YechiamW3, line 1: If you want, you can wrap these two components (if they have similar semantic meaning) in a wrapper component with a `variant` value to choose between the two

cv_next/components/cvPerview.tsx
```diff
@@ -0,0 +1,21 @@
...
+  let presentedURL = document_link;
+
+  const transformedURL = transformGoogleViewOnlyUrl(document_link);
+  if (transformedURL != "") {
+    presentedURL = transformedURL;
+  }
+
+  return (
+    <div className="w-full">
+      <iframe src={`${presentedURL}&rm=minimal`} width="1000" height="800" />
```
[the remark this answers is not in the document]
YechiamW3: Listen to CodeQL. CodeQL is good.
https://codeql.github.com/codeql-query-help/javascript/js-xss-through-dom/

cv_next/components/ui/popupButtom.tsx
```diff
@@ -0,0 +1,45 @@
...
+  children: ReactNode;
+  clickable?: ReactNode;
+  disableButton?: boolean;
+  onClose?: () => void;
+}
+
+export default function PopupWrapper({
+  children,
+  clickable,
+  disableButton,
```
YechiamW3, line 15: I don't really get this `disableButton` prop. Why not simply use the `disabled` attribute ok a `<button>` element?

cv_next/components/cvPerview.tsx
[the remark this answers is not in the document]
Aradon267: Updated it so that if the link isnt according to regex it popup that the url doesn't match the regex

cv_next/app/feed/components/inputbar.tsx
PeroG28: I prefer 2 different components, it is more readable, maintainable, and is more clean.
there is principle but i don't remember its name

cv_next/components/ui/popupButtom.tsx
PeroG28: its for disabling the button, i don't understand

Aradon267: Do you mean that instead of a div with onClick that is "defended" using the attribute, use the attribute for disabled?

YechiamW3: I'm not sure I understand this component. Are you wrapping a button (that you're expecting that get sent as prop) and override its functionality with the open/close mechanism you're doing here?
A. Again, why not use the `disabled` attribute of a button? (yes Arad, you understood correctly)
B. If you want to create a generic popup, why not simply add this button functionality as part of the popup code instead of passing it through prop?

Aradon267: If I understand correctly, this is a combination of the popup and the button that opens it
About A it'll be fixed, about B I think that the point is to allow anything we want to serve as the "popup trigger"

Aradon267: Uploaded the fix mentioned in A

YechiamW3 (approved):

PR #88, opened 2024-07-27
Cv page comment btn

Aradon267:


cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -28,69 +29,71 @@ export default function Comment({
...
+  const router = useRouter();
+  const [showAlert, setShowAlert] = useState<boolean>(false);
   const onAlertClick = async (type: boolean) => {
-    if (type) await deleteCommentAction()
-    setShowAlert(false)
-  }
+    if (type) await deleteCommentAction();
+    setShowAlert(false);
+  };
+  const supabase = useSupabase();
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:50]: hooks should be at the start of the component

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -28,69 +29,71 @@ export default function Comment({
...
 
   const addNewCommentClickEvent = async () => {
-    if (!userId) throw new Error("User not found")
-    const commentToAdd: NewCommentModel = {
-      data: commentOnComment,
-      document_id: cv.id,
-      parent_comment_Id: comment.id,
-      user_id: userId,
-    }
+    const userId = (await supabase.auth.getUser()).data.user?.id;
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:53]: We want to go through auth? Don't we want to call `getUser` api in hopes it's cached and doesn't need to go through supabase?

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -28,69 +29,71 @@ export default function Comment({
...
+    if (!userId) router.push("/inactive");
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:54]: Perhaps to `/login` instead?
Also good practice would be to redirect with a return url so that the user would authenticate then come back to the same page, back we can handle this later.

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -28,69 +29,71 @@ export default function Comment({
...
-            .get(comment.id)
-            ?.concat([commentToAdd]) as Array<any>
-        )
-        return newCommentsOfComments
-      })
-      setCommentOnCommentStatus(false)
-    })
-    mutate(comment.document_id)
-  }
+      await addComment(commentToAdd).finally(() => {
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:63]: Instead of the `newCommentsOfComments` (ofCommentsOfCommentsOfComments....) fiasco, we can try utilizing swr's [optimistic update](https://swr.vercel.app/docs/mutation#optimistic-updates) and rollback if needed.

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -106,159 +109,180 @@ export default function Comment({
...
-                  style={{
-                    transform: "translateY(2px)",
-                  }}
-                  fontSize="1.4rem"
-                />
-              </button>
-            )}
-          </span>
+  const userVoted = comment.upvotes?.includes(userId);
+  const votingSection = userVoted ? (
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:113]: Can you extract all of these small components to be outside of this big component? It's perfectly fine to keep them in the same file, but put them outside of the main component's code.

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -28,69 +29,71 @@ export default function Comment({
   commentsOfComment = [],
   setCommentsOfComments,
 }: {
-  comment: CommentModel
-  userId: string
-  childDepth?: number
-  cv: CvModel
-  commentsOfComment: any[]
+  comment: CommentModel;
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:32]: This is a big prop list, please extract to interface (doesn't need to be exposed as public)

cv_next/app/cv/[cvId]/components/commentSection/commentForm.tsx
```diff
@@ -14,15 +14,15 @@ export default function CommentForm({ cv }: { cv: CvModel }) {
...
-    const userId = (await supabase.auth.getUser()).data.user?.id
-    if (!userId) throw new Error("User not found") // TODO: handle this
-    const comment: NewCommentModel = {
-      data: formData.get("comment") as string,
-      document_id: cv.id,
-      parent_comment_Id: null,
-      user_id: userId,
-    }
+    const userId = (await supabase.auth.getUser()).data.user?.id;
+    if (!userId) router.push("/inactive");
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/commentForm.tsx:18]: Same here like the comment on `comment.tsx`.

cv_next/app/cv/[cvId]/components/commentSection/commentsSection.tsx
```diff
@@ -8,30 +8,34 @@ import { useSupabase } from "@/hooks/supabase";
...
+  const router = useRouter();
+  const { data: comments } = useSWR(cv.id, fetchCvComments);
+  const supabase = useSupabase();
+  const [userId, setUserId] = useState<string>("");
   const [commentsOfComments, setCommentsOfComments] = useState<
     Map<string, any[]>
-  >(new Map<string, any[]>())
+  >(new Map<string, any[]>());
 
   useEffect(() => {
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/commentsSection.tsx:19]: I think this should be a useMemo

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:53]: Like the actual route itself? In theory we could, another option is to use getSession to get the data from the cookies

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:54]: The login page in this version doesn't have the next param(it's with the SSR), but will add it as a TODO

Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx]: I uploaded a fix for that in the SSR branch as this one doesn;t have the next page param

cv_next/app/cv/[cvId]/components/commentSection/commentForm.tsx
Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/commentForm.tsx]: I uploaded a fix for that in the SSR branch as this one doesn;t have the next page param

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx]: Apparently we already get the user's id as a prop so this call is redundant... removed it completly

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -1,78 +1,248 @@
...
 import Link from "next/link";
-import { useSupabase } from "@/hooks/supabase";
 
-export default function Comment({
-  comment,
+const NewCommentBlock = ({
+  commentOnCommentStatus,
+  setCommentOnComment,
+  addNewCommentClickEvent,
+}: any) => {
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:24]: I know it's a bit tiresome, but please don't leave `any` type.
You can utilize the `CommentProps` interface with `Omit<>`.

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -1,78 +1,248 @@
...
-  const router = useRouter();
   const [showAlert, setShowAlert] = useState<boolean>(false);
   const onAlertClick = async (type: boolean) => {
     if (type) await deleteCommentAction();
     setShowAlert(false);
   };
-  const supabase = useSupabase();
+  const { mutate } = useSWRConfig();
 
   const addNewCommentClickEvent = async () => {
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:198]: Please remember that this _entire function_ might get recreated on every render cycle (i.e. view interaction) if not memoized; please use `useMemo` or `useCallback` (whichever you think is more appropriate).

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -1,78 +1,248 @@
...
+          // Type-checking for error
+          if (error instanceof Error) {
+            return error.name !== "AbortError";
+          }
+          // Rollback if error is not an instance of Error
+          return true;
+        },
       });
+      await addComment(commentToAdd);
+      mutate(comment.document_id);
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:244]: Why are you mutating twice? Is this how it's suppose to be handled? (I don't remember exactly the docs)

Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:244]: First is to instantly show the data to the user, second is to make sure the cache is synced with the server's state

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:198]: useCallback is for not recreating function so I went with that

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (YechiamW3) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:318]: So I believe according to docs that you should send the `addComment` function to the `mutate`, and it'll do the sync after the comment addition whilst optimistically updating the ui. Check the example here https://swr.vercel.app/docs/mutation#optimistic-updates. Also you might want to utilize (maybe in other places, not sure if here is needed) [manually updating the cache] (https://swr.vercel.app/docs/mutation#optimistic-updates).

Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx]: Took some time, but uploaded a fixed version

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -303,23 +305,27 @@ export default function Comment({
...
+              return error.name !== "AbortError";
+            }
+            return true;
+          },
+          revalidate: true,
+        }
+      );
     } catch (error) {
-      mutate(comment.document_id);
+      router.push("/inactive");
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:326]: I think we shouldn't redirect to "inactive" here, we want to do some app-level error handling if we suspect inactive user (practically: create custom error that you'll throw e.g. here and catch it at the top level).

Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:326]: If I remember correctly we decided to redirect to the login page with a next param but we just don't have it implemented here(we have it in the SSR branch that also fixed some problems with the flow of authorization itself). I'll add a TODO on relevant places.
Should I still do a custom error that'll catch this and add the TODO there?

Comment (YechiamW3) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx]: Nah it's fine leave it as TODO and fix it in a later branch

YechiamW3 (approved):

Issue #97, opened 2024-08-07
BUG: New user auth request sent every time navbar is opened


YechiamW3:
In the navbar behavior we restart the "signed in" value each time (it's saved as a local state). We need to save the sign-in data as part of a more global state (swr cache/react context) and use it as the reference point for the sign in state.
See network requests for example (should be noted: it seems swr does give back a cached data, I don't see a request on the SP logs):
![image](https://github.com/user-attachments/assets/f78d9683-faee-4223-be85-641c3fe3b483)

Relevant code:
https://github.com/magshimim-next/cv-next/blob/3759cba9857a04fa9207d29e2fc420e5003aefe9/cv_next/components/layout/navbar/popupToggle.tsx#L16
https://github.com/magshimim-next/cv-next/blob/3759cba9857a04fa9207d29e2fc420e5003aefe9/cv_next/components/layout/navbar/popupToggle.tsx#L40

YechiamW3:
Note: it only happens on small-size screens (different nabvar behavior).

PR #99, opened 2024-08-08
98 UI show loginsign out based on logged in state

YechiamW3:
NOTE: not tested locally, should theoretically work.
Added a few TODOs for your considerations (@Aradon267 welcome to check).


cv_next/hooks/useUser.ts
```diff
@@ -0,0 +1,19 @@
+import { useSupabase } from "./supabase";
+import useSWR from "swr";
+
+
+export const useUser = () => {
+    const supabase = useSupabase();
+    const { data, mutate, error } = useSWR('user', supabase.auth.getUser);
+
+    const loading = !data && !error;
+    const loginState = data && !error;
```
Aradon267, line 10: The combination of SSR and auth helpers messes this up... loginState is always undefined which means showing signout for signed out users

YechiamW3: Say, why won't it work here but will work e.g. at https://github.com/magshimim-next/cv-next/blob/0b08f2b5ed1a7a4daa5193bd62e78176770605ab/cv_next/app/cv/%5BcvId%5D/components/commentSection/commentsSection.tsx#L19
(comment section's getUser effect)?

Aradon267: No entirely sure, might be due to the await? getUser gets the response from supabase so when I print data I see `TypeError: Cannot read properties of undefined (reading '_getUser')
`

YechiamW3: OK solved a few quirks here, note:
1. `useUser` hook now gives user data as well as use session
2. the hook uses a server action that encapsulates `supabase.auth.getUser`
3. For an example (there might be more places that uses it), I've replaced existing code that controlled the user state locally with the new hook
4. Please see that the swr doesn't work exactly as I thought - it automatically revalidates many times (by default, on every mount of the component that uses it). I've disabled most of the default automatic revalidations that swr provides, and the way to revalidate it is to manually use the `mutateUser` exposed from the hook.

YechiamW3: Actually now that I think about it, for client-side usage `supabase.auth.getSession` should be enough, no? The problem is with server-side right?
So I can replace the `supabase.auth.getUser` call (again encapsulated as server action) with `supabase.auth.getSession` and keep the logic in the browser without the need to do another server-client hop.
Do you agree? (have we talked about this already? exams mess with my memory)

Aradon267: The approach itself is really good, will be useful as I think this also fixes #97 as the data is already managed using swr
Client-side calls to getSession are enough if we don't use it for any sensitive data as it can be tampered with quite easily. From the docs
`getUser Should always be used when checking for user authorization on the server. On the client, you can instead use getSession().session.user for faster results. getSession is insecure on the server.`
So if we just use it to get the name and photo that's fine, but we can't use that as user id in commenting etc

YechiamW3: So essentially there are two distinct actions here we need to differentiate:
1. data retrieval: user data, cvs, comments etc
2. data manipulation: comment post/edit, cv post, user data edit etc
The first kind can be handled in client-side supabase access, while the latter needs to be handled on the server with user validation (using `auth.getUser`). This can (should?) also be what differentiate between server actions and route handlers/client fetch.
Anyway, for now I'll simply replace the logic I've done for the user session with `getSession`, and we'll think later on how to optimize the other client/server sb separation.

YechiamW3: Actually nevermind let's keep it like that, their warnings are too scary :)

Aradon267: Yeah keeping it like that is fine
And yes the second is important that is managed on the server using getUser as you can see from #107 #108 that bypasses this
This also fixed the other issue which is great

Aradon267 (approved):

YechiamW3:
@Aradon267 I do think however (unrelated to this PR specifically) that we shouldn't pass all the user data to the client (currently we're passing the user's db data, and this PR also provides the user's session data). For user data - it's easy, let's return only necessary data; for user session perhaps the `getSession` can help, but I'm not exactly sure. Let's think about it later.

YechiamW3:
OK @Aradon267 I've thought about it a bit - we shouldn't rely on any user-related data that is passed from the browser; I've only made our already-problematic behavior more explicit but exposing all the user session.
The problem is that we should always rely (at least for user-specific actions) on user information fetched on the server (as we've talked about multiple times), and what we had prior to this PR is we fetched that data from the server to the browser and then relied on it (e.g. `userId`) with requests from the browser to the server - which means if I know user ids I can easily fetch their information (and perhaps do actions on their behalf; I didn't check all the actions we currently have).

I adjusted the `useUser` to not fetch the user session and then the user data, but instead adjusted (or more specifically added) a `getUser` function on the server which takes fresh user session from the server (via `auth.getUser`) and calls the user data fetch from there, and only that is exposed to the browser. This means that `getUserModel` should be deleted and anything that relies on it should instead rely on `useUser` (I've added a TODO comment there, let's handle it in a different PR), because it receives a `userId` as param which is sent from the browser.

Aradon267:
What about cases where I need the basic object of another user? i.e. name for the cv preview, work categories and status for the profile, etc.
There is RLS on the profiles which means that only active users can get the data of other users and that data is only name, image, work categories, and work status. The "sensitive" data we get from google isnt exposed to other users

YechiamW3:
> What about cases where I need the basic object of another user? i.e. name for the cv preview, work categories and status for the profile, etc. There is RLS on the profiles which means that only active users can get the data of other users and that data is only name, image, work categories, and work status. The "sensitive" data we get from google isnt exposed to other users

I easily fetched user data of other users by editing the user id sent from the browser.
User data that is related to the object fetched, e.g. cv object, comment object - the relevant user data that is related to these objects should be fetched along with these objects.

Aradon267:
For cvs and comments we already handled that(in most parts) so it shouldn't be that hard to combine the queries there but what about the user's profile page?

Aradon267:
So if I understand correctly, we want to only use it on the server, not completely remove it(it being `getUserModel`), and only use 
`useUser` on the browser

YechiamW3:
> So if I understand correctly, we want to only use it on the server, not completely remove it(it being `getUserModel`), and only use `useUser` on the browser, correct?

Yeah  we could do that. If we do it we need to make sure tho that we don't accidentally expose `getUserModel`.
I don't think it's something we can enforce automatically (except perhaps add a custom lint rule, which is an overkill), so we'd need to make sure to not allow such usage.

Aradon267:
After this change, it is used only on the profile(server side). We also have a getUserById from users.ts that is the same and is used to get the name of the CV uploader... The latter can be solved by using inner joins as we did for the cvData on the cv page
We can be more wary of it and add a comment that this should only be used in the server or something.

Aradon267:
I'll add the inner joins as an issue so that it won't be used, but will be there in case we need it
Feel free to merge

PR #130, opened 2024-08-23
Fix signout route

Aradon267:


cv_next/app/api/users/revalidateSignout/reoute.ts
YechiamW3, line 1: API naming is unconventional, suggestions: `revalidate`, `revalidate-signout`.
Also tyopo `reoute`. And do you really need 2 nested folders? Isn't `revalidate` enough (instead of `users/revalidate`)?

cv_next/components/layout/navbar/popup.tsx
```diff
@@ -78,18 +79,23 @@ const UserDataComponent: React.FC<{
...
       dialogRef.current.show();
     }
   }, []);
 
-  const handleSelection = (route: string) => {
+  const handleSelection = async (route: string) => {
     if (route === "Signout") {
-      supabase.auth.signOut();
+      await supabase.auth.signOut();
+      await fetchFromApi(
```
YechiamW3, line 93: Actually I think this is handled in the new `useUser` hook, before merging this please check the behavior in #99

Aradon267: The signout must happen using a supabase client component to delete the cookies automatically
Do you mean the API call/updateSignIn? because yeah maybe we could signout and manually revalidate

YechiamW3: I meant that after signing-out and re-rendering homepage, the `useUser` hook mutates and calls the revalidation itself. That's at least how I remember it worked for me locally, so please checkout the #99 branch and see if that's the behavior you expect.

YechiamW3: Ah I now see, in the #99 branch I added a call to `mutateUser` from the `useUser` hook. I believe it's better that way because we keep the SWR cache synced with our actions.
In general - I think every user-related action should be accessed/operated through the `useUser` now, so that the SWR will be up-to-date.
In merge conflicts in #99 I'll keep the `mutateUser` call, after merging it feel free to adjust if needed.

Aradon267: That's odd, currently in prod without the revalidate we must refresh after the signout
Maybe because the forced revalidate is missing or something?(like the home page isn't re-rendered)

YechiamW3: Prod includes #99?

Aradon267: Yes, was probably merged when the dev pipeline was created(when we synced the branches)

Aradon267: My mistake: in prod it does remove your data but the popup doesn't change until refresh(or a manual revalidation is needed)

YechiamW3: Hmm.. Weird. I'll check it tomorrow, bc that PR included force revalidation on the signout route

YechiamW3: OK I see the issue, it's an issue of the #99 code not of the signout page. I think it'd be better to open a PR for that instead (I'll do it). The issue is that I've introduced retrying which meant apparently that on error we keep the data we currently have and never replace it with empty data.

YechiamW3: Please check #114

Aradon267: I see
Well, the changes in #155 good seem and did fix the issue with the manual mutation on signout which was what I aimed for here so should I merge that and close this one?

YechiamW3: Yeah that's what I meant. Thanks

Aradon267:
Fixed with #155

PR #131, opened 2024-08-23
102 security check csrf account takeover

Aradon267:


cv_next/app/actions/cvs/uploadCv.ts
```diff
@@ -4,20 +4,25 @@ import { uploadCV, getCvsByUserId } from "@/server/api/cvs";
...
   cvData: InputValues;
-  userId: string;
 }): Promise<string | null> => {
+  const supabase = SupabaseHelper.getSupabaseInstance();
+  const connectedUser = await supabase.auth.getUser();
+  if (connectedUser.error || !connectedUser.data.user) {
+    logger.error("Error getting user");
+    return "Error getting user.";
+  }
+  const userId = connectedUser.data.user.id;
```
YechiamW3, line 25: I'll approve for now but add a TODO to change this to the new `getUser` function in #99

cv_next/app/auth/callback/route.ts
```diff
@@ -2,19 +2,27 @@
...
     if (!error) {
-      return NextResponse.redirect(
-        `${origin}` + Definitions.AUTH_DEFAULT_REDIRECT + next
-      );
+      if (checkRedirect(next)) {
+        return NextResponse.redirect(
+          `${origin}${Definitions.AUTH_DEFAULT_REDIRECT}${next}`
+        );
+      } else {
+        return NextResponse.json(
```
YechiamW3, line 21: Do we handle this somehow? Does this cause a redirect to some error page we have?

cv_next/app/signout/page.tsx
```diff
@@ -1,10 +1,10 @@
 "use client";
-import { useSupabase } from "@/hooks/supabase";
+import { createClientComponent } from "@/helpers/supabaseBrowserHelper";
 import { useRouter } from "next/navigation";
 
 export default function Page() {
   const router = useRouter();
-  const supabase = useSupabase();
+  const supabase = createClientComponent();
```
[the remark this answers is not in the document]
Aradon267: useSupabase uses auth-helpers which are problematic when used with SSR. createClientComponent is using the SSR version of creating a supabase client for client-side things(like signing out)

cv_next/app/auth/callback/route.ts
Aradon267: Redirects to a black page with the text
In retrospect a bad idea - this is called only when the user deliberately tries accessing a non existing page after login so maybe we redirect to feed or 404 or something?

YechiamW3: 404 is good

cv_next/app/signout/page.tsx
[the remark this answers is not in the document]
YechiamW3: OK I think it's something that needs to be handled nextjs-side, but I'm fine with this fix and will think about it later.

YechiamW3 (approved):

PR #132, opened 2024-08-23
BUG: username not showing in preview

Aradon267:


YechiamW3 (approved):

YechiamW3:
Approved of course, but do we have usernames?
That's actually something I wanted to talk to you about, if we opt to using usernames (even auto-generated ones based on fullname), that'd be much better than using ids for apis (and tbh especially for the profile url).

Aradon267:
Usernames are targeted mainly at people who don't have their name on their email but want to let others know who they are(for example Assaf is called Tom Smith for some reason so he could add a username that will be used instead)
You mean adding another "ID" field?

PR #134, opened 2024-08-23
100 fix npm build errors on dev

Aradon267:


cv_next/app/login/page.tsx
```diff
@@ -1,27 +1,13 @@
 "use client";
```
YechiamW3, line 1: fyi can delete thtis

YechiamW3 (approved):

PR #151, opened 2024-08-29
111 UI when the user hasnt uploaded a cv yet the user page is not centered

Aradon267:
image


cv_next/app/auth/callback/route.ts
```diff
@@ -16,10 +16,13 @@ export async function GET(request: Request) {
...
         return NextResponse.redirect(
-          `${origin}${Definitions.AUTH_DEFAULT_REDIRECT}${next}`
+          `${process.env.NEXT_PUBLIC_BASE_URL}${Definitions.AUTH_DEFAULT_REDIRECT}${next}`
         );
       } else {
-        const notFoundUrl = new URL("/not-found", origin);
+        const notFoundUrl = new URL(
+          "/not-found",
+          process.env.NEXT_PUBLIC_BASE_URL
+        );
```
YechiamW3, line 25: Is this change related to this PR?
Also - I recall a similar change in a different PR, please check.

Aradon267: Isn't related, I'll check where it was added because I think that it was

YechiamW3 (approved):

TheGreenOak (approved):

PR #162, opened 2024-09-02
removed the feed button from the user popup

TheGreenOak:
it looks weird but the whole popup needs to move and be more like an actual popup and less like a modal

### DIFF
--- a/cv_next/components/layout/navbar/popup.tsx
+++ b/cv_next/components/layout/navbar/popup.tsx
@@ -13,11 +13,6 @@ const navLinks = [
     path: "/login",
     req_login: false,
   },
-  // {
-  //   route: "Feed",
-  //   path: "/feed",
-  //   req_login: true,
-  // },
   {
     route: "Upload",
     path: "/upload",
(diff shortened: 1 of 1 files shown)


cv_next/components/layout/navbar/popup.tsx
```diff
@@ -13,11 +13,11 @@ const navLinks = [
...
-  {
-    route: "Feed",
-    path: "/feed",
-    req_login: true,
-  },
+  // {
+  //   route: "Feed",
+  //   path: "/feed",
+  //   req_login: true,
+  // },
```
Aradon267, line 20: We can just delete these as we don't want them anymore

Aradon267, line 20: We can just delete these as we don't want them anymore

Aradon267 (changes requested):

cv_next/components/layout/navbar/popup.tsx
Aradon267: @TheGreenOak reminding you this

YechiamW3 (approved):

TheGreenOak (commented):
now its removed instead of commented

TheGreenOak:
@Aradon267 why is the merging still blocked?

Aradon267 (approved):

Aradon267:
I think that it was because  I also needed to approve this...
Can be merged now!

PR #163, opened 2024-09-06
82 website documentation

Aradon267:
Updated the documentation as well as some other things(contrib and drawio)
The contribute assumes that we decide to not allow any contribs from others so its just a draft
There are also some TODOs


README.md
```diff
@@ -1,22 +1,41 @@
...
 git clone https://github.com/magshimim-next/cv-next.git
-```
-```
+cd cv_next
 npm install
 ```
-then fill in the firebase config file with the credentials
 
-## Usage
+then fill in an .env.local config file with the credentials to supabase(supabase url and anon key), and add the base URL(usually localhost for local development) as seen in the [example env file](./cv_next/.env.local.example).
```
YechiamW3, line 20: Add to contact us on the community chat for configuration

docs/DESIGN.md
```diff
@@ -44,23 +72,112 @@ We will also provide a general comment section under each post.
...
 ### Users
-|field|type|
-|:---|:---|
-|created|datetime|
-| lastLogin|datetime|
-|email|string|
-|name|string|
-|userTypeID|number|
+
+As this table is auto generated by Supabase, we only use several fields from it.
```
YechiamW3, line 91: I don't really care that we put it here (it can be easily viewed in the code), but do we really want to? Putting it here means we need to manually update this on every schema change.. (this comment is relevant to everything under Schemas & Collections.

CONTRIBUTING.md
```diff
@@ -0,0 +1,35 @@
...
+
+TODO: decide how we want to accept feedback and more suggestions
+
+### How to start
+
+Grab an issue you think you can manage, assign it to yourself, and create a fork for it!
+
+### Conventions
+
+TODO: enter conventions
```
Aradon267, line 35: @YechiamW3 should we add something here or just delete it?

CONTRIBUTING.md
```diff
@@ -0,0 +1,35 @@
...
+
+## For members
+
+Thank you for wanting to contribute to the project and the community!
+
+First, if you have any question, contact us from the community group and we'll gladly help!
+
+### Feedback and Suggestions
+
+TODO: decide how we want to accept feedback and more suggestions
```
Aradon267, line 27: @DenisGran should we point them to the community chat or something?

CONTRIBUTING.md
YechiamW3: If we want to write conventions we'd probably want a separate file; for now I think we should keep it empty and add later if we see necessary because keeping the conventions in check via the CRs is good enough for now, in my opinion.
Anyway, you can assign to me an issue to write such a file.

Aradon267: Added #177

Aradon267: Resolving this but updated the todo

CONTRIBUTING.md
Aradon267: Updated to creating issues or via a link in the community

PR #90, opened 2024-08-03
Link query real

nevogithub:
Pull request for category links

PR #167, opened 2024-09-13
Link query new

Aradon267:
This is the combination of #90 and #65 as they were picked up by me.
I made sure that the threads from there were implemented here.


cv_next/app/feed/components/filterPanel.tsx
```diff
@@ -16,14 +19,40 @@ export const FilterPanel = ({
...
+    }
+
+    if (searchValue) {
+      params.delete("description");
+      params.append("description", searchValue);
+      router.replace(`${pathname}?${params}`);
+    } else {
+      params.delete("description");
+      router.replace(`${pathname}?${params}`);
+    }
```
YechiamW3, line 49: This handles two separate effects, it should be separated to two different `useEffect`s.

cv_next/server/api/cvs.ts
```diff
@@ -156,34 +156,70 @@ export async function getPaginatedCvs(
 
     const supabase = SupabaseHelper.getSupabaseInstance();
     let query = supabase
-      .from(Tables.cvs)
+      .from("cvs")
```
YechiamW3, line 159: Why make this change?

cv_next/server/api/cvs.ts
```diff
@@ -156,34 +156,70 @@ export async function getPaginatedCvs(
...
       .select("*")
-      .range(from, to - 1);
+      .order(CvKeys.created_at, { ascending: false });
+    let profileQuery = supabase.from("profiles").select("id");
```
YechiamW3, line 162: And use `Tables.profiles`, etc here too

cv_next/server/api/cvs.ts
```diff
@@ -156,34 +156,70 @@ export async function getPaginatedCvs(
...
 
     logger.debug(filters, "filters");
-    if (filters) {
-      if (filters.searchValue) {
-        query = query.textSearch(CvKeys.description, filters.searchValue);
-      }
-      if (filters.categoryIds) {
-        logger.debug(filters.categoryIds, "category ids");
-        query = query.overlaps(CvKeys.cv_categories, filters.categoryIds);
+
```
YechiamW3, line 165: Everything in this function that depends on the `searchValue` is complicated to read. There are multiple `if (filters?.searchValue)` statements, why? It'd be more readable to separate these statements to local functions with appropriate names

cv_next/app/feed/components/feed.tsx
```diff
@@ -64,6 +71,30 @@ export default function Feed() {
...
+      } else {
+        filters.categoryIds = null;
+      }
+
+      filters.categoryIds?.filter(function (element) {
+        return element !== undefined;
+      });
+      if (filters.categoryIds?.length && filters.categoryIds[0] === undefined)
+        filters.categoryIds = [];
+
```
YechiamW3, line 97: Can this be extracted to a separate function?

cv_next/server/api/cvs.ts
Aradon267: No idea tbh

cv_next/app/feed/components/filterPanel.tsx
Aradon267: Won't that cause an override of other values? Like if I only enter the searchValue what should I put in the categories?

YechiamW3: I might not understand either the business logic or the logic written here correctly. What happens here, if I understand it correctly, we re-route the feed by the search params, in a cascading way of categories -> description search. Is this the intended behavior? Am I understanding it incorrectly?

Aradon267: Technically could be interchangeable(I don't really care if its categories or description first)

YechiamW3: What is this effect supposed to do?
`router.replace(`${pathname}?${params}`);` doesn't redirects?
I might need an example of a use-case. I get e.g. `/feed?category="abc"&description="def"`. In this code, we remove the `category` param, add the `category` param with the changed values and then redirect. Then we hit this component again with the same uri.. Will we ever get to the `searchValue` check?
Also, why do all the `delete` and `append` instead of `set`?

YechiamW3: Parameters changes means uri change means re-render, essentially. From the docs about `replace`: "Perform a client-side navigation to the provided route without adding a new entry into the browser’s history stack."
So a client-side navigation occurs, meaning a re-render. It might not look like a re-render because it's the same component that only gets its props changed, meaning a prop diff is being made.
The main question is, using this method, do you ever get `searchValue` change have any effect? Especially when you also have a `category` param set?
And I'm not sure why would the parsing be any different via `set`, the code here basically does a `set` but in a more complicated way..

Also btw in order to _not_ cause a re-render, there's a [github discussion](https://github.com/vercel/next.js/discussions/18072#discussioncomment-109059) which suggest accessing the `window` directly, which I don't really like unless we absolutely have to. I think we can solve it better here - there's no problem in re-rendering, you should just be aware of it as it will most likely cause part of this code currently to misbehave.

Aradon267: I mean yeah when I edit `searchValue` the URL changes and the filter is applied
With `set` it does `category=categoryAcategoryB` according to a string and we parse with `getAll` and `category=categoryA&category=categoryB`

So if we want to avoid accessing `window`, is it better to leave the rerenders like that?

YechiamW3: Wait, what you want is `categeory=categoryA&category=categoryB`? Or `categeory=categoryA,categoryB`?
I personally prefer the latter, which can be handled with e.g:
```params.set("category", categoryIds);```

And it's kinda weird that `router.replace` doesn't stop the useeffect flow, but it might be because it only takes effect at the end of the render cycle.
Anyway, regarding separating the effect - bear in mind that changes in only one param will cause "changes" (i.e. code execution, even if the final state will look similar) to both parameters.

Full code, in my opinion, should look like:

```
useEffect(() => {
    const params = new URLSearchParams(searchParams);
    params.set("category", categoryIds);
    router.replace(`${pathname}?${params}`);
    onChange({
      categoryIds: categoryIds
    });
  }, [categoryIds, onChange, searchParams, router, pathname]);

useEffect(() => {
    const params = new URLSearchParams(searchParams);
    params.set("description", searchValue);
    router.replace(`${pathname}?${params}`);
    onChange({
      searchValue: searchValue
    });
  }, [searchValue, onChange, searchParams, router, pathname]);
```

If you want, it can also be:

```
useEffect(() => {
    const params = new URLSearchParams(searchParams);
    params.set("category", categoryIds);
    params.set("description", searchValue);
    router.replace(`${pathname}?${params}`);
    onChange({
      categoryIds: categoryIds,
      searchValue: searchValue
    });
  }, [searchValue, categoryIds, onChange, searchParams, router, pathname]);
```

(you can keep the param deletion if you want, I just think it's unnecessary)

Comment (Aradon267) [reply cv_next/app/feed/components/filterPanel.tsx]: What we currently have is option A `categeory=categoryA&category=categoryB`
I tried separating similarly to how you did(just with option A in mind) but the `onChange` expects a valid structure with both values so I separated to different effects but still both use that `onChange`

Comment (Aradon267) [reply cv_next/app/feed/components/filterPanel.tsx]: Actually after splitting into 2 effects the description isn't being set at all...

Comment (YechiamW3) [reply cv_next/app/feed/components/filterPanel.tsx]: I understand the problem, it's to do with the `onChange` function. I've checked the branch locally to understand better what exactly is being done here; we basically have two types of states that control the same data, which is not the best approach (and actually hinders "single source of truth"). It'd be better to rely only on the search params as state, and at most mirror it to a single state object (just for the sake of better data control in the code).
I have some changes to reflect what I mean; I can push it here if you want.
By the way, I've noticed that the router change takes a bit to load (especially the first one), and I assume it's because of the massive network load because of all the requests to feed. I never looked into it, are there more requests than needed, or is there just way too much data in the dev environment?

Comment (Aradon267) [reply cv_next/app/feed/components/filterPanel.tsx]: If you can push that would be great
The feed makes several requests at a time for each CV(and we also need to remember that we need to fetch the cvs):
1. Blurred preview
2. Updated(and updating) preview
3. CV's uploader
There are caching and memos which I think work as intended but not quite sure...

Comment (YechiamW3) [reply cv_next/app/feed/components/filterPanel.tsx]: I need to look into the caching mechanisms used, because I don't think they work properly (tho it might be simply dev env-related non-optimization, but I doubt it).
And what do you mean by "updated preview"? and the CV uploader?
And I'll push soon^tm

Comment (YechiamW3) [reply cv_next/app/feed/components/filterPanel.tsx]: Pushed the commit. Please look through it - it is NOT complete; if you think it should be changed (or not accepted), just write comments on it. I've also added two notes on (what I think are) necessary improvement performance-wise.
Note that this page is the heaviest performance-wise of the site, we need to make sure we keep it as performant as we can.

Comment (Aradon267) [reply cv_next/app/feed/components/filterPanel.tsx]: I saw the commit, and it looks good and it's working as intended
Regarding the notes, the debounce would be important, and we can change the formatting if this makes parsing easier.

Do you want me to take this or do you already have a general idea of how things would look like?

Comment (YechiamW3) [reply cv_next/app/feed/components/filterPanel.tsx]: Planned on letting you handle it, but I just finished my last exam so I have a couple of days to work on the project so I can take it :)

File: cv_next/server/api/cvs.ts [hunk 6]
```diff
@@ -158,38 +158,77 @@ export async function getPaginatedCvs(
...
     if (error) {
       logger.error(error, "getPaginatedCvs");
       return null;
     }
 
-    return { page: page, cvs: cvs as CvModel[] };
+    const uniqueCvs = [...new Map(cvs.map((cv) => [cv.id, cv])).values()].slice(
+      from,
+      to
+    );
```
Aradon267, line 185: Isn't it expensive to always get everything from the DB and just return a specific amount? Or does that even out(or even outweigh) the price of calling all the filters on just a small amount?

cv_next/app/feed/components/filterPanel.tsx
Aradon267: Ran the changes locally, and they all worked perfectly, turning debounce to a hook was a nice addition

Aradon267:
lmk what you think about the threads, I'll try to change things accordingly before the weekend

cv_next/server/api/cvs.ts
YechiamW3: Wait wait I didn't notice this code.
Why did we deleted the `range` in the supabase query?
The paginated selection should be in the db, not manually after getting all the results..

cv_next/server/api/users.ts
```diff
@@ -30,6 +31,30 @@ export async function getUserById(
   }
 }
 
+export async function getUserIdByName(
```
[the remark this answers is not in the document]
YechiamW3: Yeah that's a weird function. Should be deleted.
But perhaps a cleanup PR is needed? I see that this isn't related to this PR but rather from the merge, so it doesn't have to be handled in this PR.

cv_next/server/api/cvs.ts
Aradon267: That's what I thought, I'll move it back up

cv_next/server/api/users.ts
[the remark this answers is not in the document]
Aradon267: Created #193 so resolving this

cv_next/app/feed/components/feed.tsx
```diff
@@ -1,19 +1,26 @@
...
 import { useApiFetch } from "@/hooks/useAPIFetch";
 import { ScrollToTop } from "@/components/ui/scrollToTop";
+import { useSearchParams } from "next/navigation";
+import { toCategoryNumber } from "@/lib/utils";
+import { useDebounceCallback } from "@/hooks/useDebounceCallback";
 
 export default function Feed() {
+  const searchParams = useSearchParams();
+  const uriCategories = searchParams.get(CATEGORY_PARAM);
+  const description = searchParams.get(DESCRIPTION_PARAM);
```
Aradon267, line 22: Considering the fact that this is presented to the user, is there a chance that this(and possibly other user-based inputs) are prone to XSS or sqli?

YechiamW3: Hmm.. This is being treated as String and there's no `eval` on it; it's also not used directly to be sent to the server (as there's the whole string-to-id transformation), so I can't think of an obvious way this can be abused. But lmk if you think there's any reason to do some special parsing here.

YechiamW3: As regards to other user-based inputs - sure, this should be on a per-case basis how to treat the information. We can check e.g. if supabase doesn't do anything on their side to resolve sqli-type attacks (for e.g. search functionality), but because it's essentially an orm I'm pretty sure they do.

Aradon267: Ok so it's handled on their end
Resolving this

Aradon267:
@YechiamW3 fyi I merged the cleanup PR so everything should be ready

YechiamW3 (approved):

PR #170, opened 2024-09-14
136 UI feed should be the landing page

Aradon267:


YechiamW3:
Wait, but what about logged-out users? `about` seems like a weird page for logging in..

Aradon267:
Will rendering based on connectivity status(with the useUser) be better?

YechiamW3:
> Will rendering based on connectivity status(with the useUser) be better?

Yeah it sounds better, and sounds like a familiar UX (in popular sites).

cv_next/app/(landing)/page.tsx
```diff
@@ -1,7 +1,18 @@
...
+import { AboutLayout } from "@/app/about/components/aboutLayout";
+
+export default function Page() {
+  const { loading, loginState, mutateUser } = useUser();
+
+  useEffect(() => {
+    mutateUser();
+  }, [mutateUser]);
+
+  if (loading || !loginState) return <AboutLayout />;
```
YechiamW3, line 14: if `loading`? I think it should be a separate option, because the loading can take a few ms that the react cycle has already finished in, which would redirect the user (logged in but in a small delay) to the about page.

Also this can actually be entirely ssr'd, you check the login state when rendering on the server and return the correct redirection (which will probably be a better UX). And I stress - a redirection is more appropriate than a manual render of the about page in the landing page. More specifically, I think the best option (which is similar to other sites' UX) is to redirect to the feed page if logged in, and if not render here the "about" page (i.e. leave the about-like page in the landing page, and simply auto-redirect to /feed if logged in).

Aradon267: updated the landing page to use server and redirect to feed

YechiamW3 (approved):

Issue #166, opened 2024-09-10
DB: ERD and Remove unnecessary fields


lordYorden:
- [x] Organizing database and removing unnecessary fields #118 
- [x] Creating an updated ERD

lordYorden:
@Aradon267 Look at the new ERD, Is it okay to implement this?

Aradon267:
Hmmm... Where is it?

Aradon267:
Saw the changes, they were excellent! Several things to note and ask(mainly because I'm not sure about the signs like double circles I guess are an array?):
1. in CV, CategoryID is an array(as each CV can have up to 3 categories) same for workCategory in Profile.
2. In CV we don't need CategoryID and Category(as mentioned earlier)
3. is CommentOn something in the comments that references a profile and a CV?

lordYorden:
The rhombuses represents relationships between entities and I suggest making a tri relation because a user comments on a CV hence tri, between profile, CV and comment

lordYorden:
As for categories there is currently a major duplication because both work and CV has the exact same category system so I suggest separating it to a completely different entity for both. Using a new table, and making the table structure a lot more moduler and forward thinking.

lordYorden:
I've added the Resulted Table Structure of the ERD also in the doc. @Aradon267

Aradon267:
Ok this looks good and will ease the queries for sure!
2 things that I'm scared of here:
1. size limitations of the database (as we talked about) - the change adds a lot of tables and I'm afraid of hitting the limit
2. integrating this change we the code

lordYorden:
As for space constraints, I will simulate the new db structure locally in order to understand the resources needed.

Issue #118, opened 2024-08-17
DB: Remove unnecessary fields from supabase


Aradon267:
The category_id of cvs and user_type of profiles aren't used.
Updates should also be made in the code

PR #181, opened 2024-09-29
DB: ERD and Remove unnecessary fields

lordYorden:
#166 finished Erd and Remove unnecessary fields from cv and profile #118


lordYorden:
fixed the problem

Aradon267:
Should we also make these changes appear in the database itself and the types file?

lordYorden:
They are in the type file

lordYorden:
As for the database it should stay null for now

lordYorden:
Maybe in the release

Aradon267:
Saw the types, you're correct, my mistake

What I mean is that in the DB it is set to allow nullable, isn't it?

lordYorden:
Correct, and that's what I said we should keep it that way for now, maybe change it in prod before release

Aradon267 (dismissed):

cv_next/app/actions/cvs/uploadCv.ts
```diff
@@ -25,10 +25,12 @@ export const checkUploadCV = async ({
     return "Error getting user.";
   }
   const userId = connectedUser.data.user.id;
+  //added a check, why undefined?
```
YechiamW3, line 28: Who are you asking? 🤔 
If this is a team-related question, ask the relevant programmer (git blame). There's no point in leaving such a comment in the code :)

cv_next/app/upload/page.tsx
```diff
@@ -47,7 +47,8 @@ const InputRow = ({
 };
 
 export default function Page() {
-  const [catagoryId, setCatagoryId] = useState<InputValues["catagoryId"]>(null);
+  const [catagoryId, setCatagoryId] =
+    useState<InputValues["cvCategories"]>(null);
```
YechiamW3, line 51: While you're at it, use constant instead of string for the field name (see relevant examples in other places in the code).
Also see other lines that needs such adjustment (53 and 54 for example).

cv_next/app/actions/cvs/uploadCv.ts
lordYorden: got it, but you also didn't answer my question

cv_next/app/upload/page.tsx
lordYorden: all I did was to change the name from catagoryId to cvCategories
and I do not understand the changes you are requesting

lordYorden: and where can i find an example

cv_next/app/actions/cvs/uploadCv.ts
lordYorden: and the git blame shows @Aradon267

docs/cv-next-ERD.drawio
YechiamW3, line 1: If we're creating ERD, can we use the proper entities objects in the diagram? E.g. like this:
image

Though it might be personal preference, I do think it's easier to navigate.

lordYorden: that is not an ERD

lordYorden: what you are suggesting is a class diagram

lordYorden: heres an example: https://vertabelo.com/blog/chen-erd-notation/

YechiamW3: What you're referring to is _one type_ of an ERD notation :)
Look at https://www.lucidchart.com/pages/er-diagrams for more types. I personally don't like Chen's notation (it's easier to have messy visuals that way, and requires a bit of a mental load to remember the exact symbolic meaning).
But it's down to personal preference, so I'm happy to let the team decide.

lordYorden:
@Aradon267 should we also change `supabase-definitions.ts`?
because @YechiamW3 isn't sure if it is generated automatically

docs/cv-next-ERD.drawio
lordYorden: ok, in any case we decided not to work on it in the foreseeable future

lordYorden: ok

lordYorden: in any case I'm opening a new Issue for the updated ERD

Aradon267:
supabase definitions isn't automatic I'm afraid so yes it should also be updated

cv_next/app/upload/page.tsx
Aradon267: look for Cvkeys or something there

lordYorden: i found exactly what he meant and apparently it not possible with an Interface such as InputValues

YechiamW3:
> @YechiamW3 what is this function in `csv.ts` called `getAllCvsByCategory` I can't find it's use

Probably a to-be-used function that was never used. You can delete it if you want (if it contains data keys you're deleting).

cv_next/app/upload/page.tsx
Aradon267: Maybe it's because of the type that is mentioned there(in the supabase definitions)

Aradon267:
Yes that function can be deleted

lordYorden:
Already deleted

cv_next/app/upload/page.tsx
lordYorden: image

Aradon267: Wouldn't doing:
```ts
  const [catagoryId, setCatagoryId] = useState<number[]>([]);
  const [description, setDescription] = useState("");
  const [link, setLink] = useState("");
```

Solve the problem? No hardcoded strings or anything(obviously a bit more changes are needed for this but generally isn't it fine?

Aradon267:
@lordYorden just to be sure, this also removes the user type from the profiles table right(not profile perms, but profiles)?

lordYorden:
Yes, Correct

YechiamW3 (approved):

PR #183, opened 2024-10-02
144 UI show the inactivity status with instructions on what to do in the landing page

Aradon267:


YechiamW3:
Wait, does this PR works nicely with #170?

Aradon267:
Need to change it to use post and not URL to prevent phishing

Aradon267:
I decided to switch to having a set amount of errors that can be shown instead of the URL things so phishing and stuff won't be an issue.

I'll merge with the dev and see how the changes work with the previously mentioned issue

Aradon267:
@YechiamW3 fyi, I've merged with the dev of #170 and it worked

Aradon267:
Is there anything else that's missing here?

cv_next/app/about/components/aboutLayout.tsx
```diff
@@ -1,17 +1,59 @@
...
+  useEffect(() => {
+    const error = searchparams.get("error");
+    if (error == Visible_Error_Messages.InactiveUser.keyword) {
+      setErrorMsg(Visible_Error_Messages.InactiveUser.title);
+      setErrorDescription(Visible_Error_Messages.InactiveUser.description);
+    } else if (error != null) {
+      setErrorMsg(Visible_Error_Messages.DefaultError.title);
+      setErrorDescription(Visible_Error_Messages.DefaultError.description);
+    }
+  }, [searchparams]);
```
YechiamW3, line 32: I think now would be a good time to create an ErrorPopup component, or more preferably ErrorContext (with a ContextProvider) that'll catch an emitted error object (containing e.g. message, description, and callback).

cv_next/app/profile/[profileId]/components/editables/editableUsername.tsx
Comment (YechiamW3) [inline cv_next/app/profile/[profileId]/components/editables/editableUsername.tsx:1]: Perhaps, in light of #85, we don't need the changes in these files? It'll just clutter the merge process.
We can merge #85 first and then add the change more simply in this PR.

Comment (Aradon267) [reply cv_next/app/profile/[profileId]/components/editables/editableUsername.tsx:1]: So should I remove these changes from this PR?

Comment (YechiamW3) [reply cv_next/app/profile/[profileId]/components/editables/editableUsername.tsx:1]: Or we just merge #85 first and then edit this PR accordingly (simply because those changes will be redundant after #85).
Whatever you prefer.

Comment (Aradon267) [reply cv_next/app/profile/[profileId]/components/editables/editableUsername.tsx:1]: Let's merge #85 and then

Comment (Aradon267) [reply cv_next/app/profile/[profileId]/components/editables/editableUsername.tsx:1]: Will handle this now that we merged #85

Comment (Aradon267) [reply cv_next/app/profile/[profileId]/components/editables/editableUsername.tsx:1]: Actually, these changes were made only in case we couldn't get the uid in the editable modes(which we now just send to the component) so we don't need these

cv_next/app/about/components/aboutLayout.tsx
Aradon267: I think I added what you meant, lmk if more changes are needed

cv_next/providers/error-provider.tsx
YechiamW3, line 1: Very nice!

YechiamW3 (approved):

PR #190, opened 2024-10-13
UPDATE: generate username based on user's name and timestamp hash

YechiamW3:
Notice the TODOs.
Lemme know if you think we should solve these TODOs in this PR, or merge it and make these adjustments later.


Aradon267:
The problem that I foresee is that I'm not sure we have a reliable way to get the generated username - we have a way to do this with the ID(getUser of Supabase) - where will this be stored?

Aradon267:
I think that both TODOs can be moved to different issues
The first login todo can be an issue regarding the first login(first users guide)
The API changes, as i've written in the issue:
The problem that I foresee is that I'm not sure we have a reliable way to get the generated username - we have a way to do this with the ID(getUser of Supabase) - where will the name be stored? User metadata is the only thing that I can think that is maybe possible

YechiamW3:
> I think that both TODOs can be moved to different issues The first login todo can be an issue regarding the first login(first users guide) The API changes, as i've written in the issue: The problem that I foresee is that I'm not sure we have a reliable way to get the generated username - we have a way to do this with the ID(getUser of Supabase) - where will the name be stored? User metadata is the only thing that I can think that is maybe possible

You mean session-wise? We can still `getUser` and validate its username against the given username, same as with validating the id.
The username is currently part of the User object.
The main point is to work *client-side* with the username, we don't really care if server-side we also use the id.

YechiamW3:
Note: I **wish** supabase would allow for atomic operations (transactions), but apparently they don't (more specifically, the underlying lib they use don't, for now). So we can't be sure the `username` won't have a possible collision, but I made sure that can only happen in quite an *extreme* edge case by moving the usernames fetching closer to the update, and making the hashing a bit more unique.

Note2: we can actually have transactions, if we do an RPC to a stored procedure that does this username generation, directly at the db level on supabase. I personally hate stored procedures, so I won't advocate for this solution, but I have to bring it up.

Note3: we can *actually* entirely solve this uniquity verification by simply making `username` a unique field; that'd simply mean it can not be NULLd, which means we either drop-table to the current Profiles table and letting users re-sign with this logic in place, or generate some random string there (which needs to be something I can differentiate from the generation I currently do, so that I'd include such cases in my to-be-replaced usernames).

Aradon267:
> The main point is to work _client-side_ with the username, we don't really care if server-side we also use the id.
Ahhh I see, ok that's the thing that I think I missed

Option 2 isn't good according to what you said(out of curiosity, why?)
In option 1 you mentioned "for now", does that mean the underlying library has intentions to support it? If so is there a solution that will make it easier to move to that in the future(as long as these plans are close) or maybe just go with option 3 to ensure uniqueness and just generate for the user

YechiamW3:
> Option 2 isn't good according to what you said(out of curiosity, why?) In option 1 you mentioned "for now", does that mean the underlying library has intentions to support it? If so is there a solution that will make it easier to move to that in the future(as long as these plans are close) or maybe just go with option 3 to ensure uniqueness and just generate for the user

Stored procedures are something that you have to maintain really heavily. It's an extra logic to your app that doesn't exists where your actual logic is. And sql, so heavily non-typed and non-debuggable.

And not really. It seems more like they want to dismiss the issue..

Aradon267:
I see
Now that I think about it, considering the fact that we will move to self hosted supabase(which I think means deleting all data), we can choose 3

YechiamW3:
> I see Now that I think about it, considering the fact that we will move to self hosted supabase(which I think means deleting all data), we can choose 3

OK this is great. We can do that. Let's think though how to resolve the edge case of 10 failed attempts. What I think is fallback to generate uuid (preferably via the pg's auto-generation), or just populate the same value of the id to the username. Will look a bit ugly, but will work, and it's a very specific edge case that has a very slim chance of occurrence. And we'll request the user to change username either way.

Aradon267 (approved):
I'll add the TODOs as new issues

PR #197, opened 2024-10-25
127 create hall of fame

Aradon267:
This is how it looks:
image

Currently, it's inaccessible(no button or anything). How should I link it? The navbar will be gone soon so maybe just a link in the footer or even in the about page?


Aradon267:
3 important things to add:
- [ ] Click to go up
- [ ] Link to feedback or something
- [ ] A link to the page somewhere

cv_next/lib/definitions.ts
```diff
@@ -93,3 +94,21 @@ export const Credits = [
   "Assaf Kabesa",
   "Alon Green",
 ];
+
+export const External_Credits = [
```
YechiamW3, line 98: Why do this manually and not, e.g. use the github api to extract the conttributors?

Aradon267: Using actions or regular API from the typescript?

Besides, some contributors were only in gitlab and this part is mainly for the current and former "core" team and not randoms

YechiamW3: Oh I see, this isn't meant to be git-related contributors (like on the main page of the github project), but simply showing legacy contributors? Like, the specific names and contributions you put are not random but are meant to be hard-coded like the core team values?

Aradon267: External are people that participated in the bug bounty and those that will help with PRs in the future as it's OS

Team are more core people or people that used to be really into the project

Aradon267:
Currently it's in the user popup just so that we won;t forget about it but we should think about integrating this in another place

Aradon267:
@DenisGran suggested that we contact the people that are there and ask them if they're fine with being there

cv_next/lib/definitions.ts
YechiamW3: I see.
Bug bounty - that surely has to be hard-coded. But "help with PRs" - that's very easily fetched via [api](https://docs.github.com/en/rest/collaborators).
If you don't want, that's absolutely fine, just thought it'd be a nice idea to prevent manual changes to things that can be automated :)

YechiamW3 (approved):

cv_next/lib/definitions.ts
Aradon267: I've added an issue for this to happen in a future version

PR #199, opened 2024-10-25
UPDATED: cookie version is overriden

Aradon267:
Before the update:
```
# npm audit report

cookie  <0.7.0
cookie accepts cookie name, path, and domain with out of bounds characters - https://github.com/advisories/GHSA-pxg6-pf52-xh8x
No fix available
node_modules/cookie
  @supabase/ssr  *
  Depends on vulnerable versions of cookie
  node_modules/@supabase/ssr

2 low severity vulnerabilities

Some issues need review, and may require choosing
a different dependency.

```

After:
```
found 0 vulnerabilities
```


YechiamW3 (approved):

YechiamW3:
General remark: we should probably start adding tests, at least to have more certainty on changes like this.

Aradon267:
Can you give an example for a test? I mean this was something that the security advisor suggested

YechiamW3:
> Can you give an example for a test? I mean this was something that the security advisor suggested

I meant generally in the code; if we'd have tests that we rely on we could simply run them and be sure that the change didn't affect anything (generally a good thing; in this particular case I simply can't really tell what the change actually does, unlike actual code change).

Aradon267:
I see. I tried the functionality afterwards and everything seemed to work correctly.
I'm not really familiar with testing websites, how do you add tests?

YechiamW3:
> I see. I tried the functionality afterwards and everything seemed to work correctly. I'm not really familiar with testing websites, how do you add tests?

Yeah that's why I only added as a general remark, and not a comment on the code :)
Testing is usually done using vitest and react-testing-library. We should probably add an issue to gradually add tests.

Issue #202, opened 2024-10-26
Update all usage of user's ID with new username


Aradon267:
You can read #190 to get a better idea of why and fallbacks and stuff

lordYorden:
I read all of #190 and I have some questions:

1. Do we change only client-side logic?
2. The fallback that is mentioned uses the UUID from before, is that the case?

Aradon267:
I believe that yes and yes(I think we decided that anything the user sees like the URL will be the username and the username would be resolved to get the UUID)

lordYorden:
@Aradon267 I tried to change the logic to use the username 
but for some reason even though the rest of the logic uses username
the page definition must use profileId still.
![image](https://github.com/user-attachments/assets/60d0d282-d382-4c26-8499-b4ddf37a5c82)

What the user now see
![image](https://github.com/user-attachments/assets/e6e78b9b-6baf-4c2c-ae9a-1b60ab52d9a0)

Aradon267:
Not entirely sure that I understand, what do you mean must use profileid? Maybe it due to te folder's name?

lordYorden:
That's what i thought but i wasn't able to change it.
resulted in an EPERM error

lordYorden:
Yes, it is working great

Aradon267:
From #190
> The main point is to work client-side with the username, we don't care if server-side we also use the id.

So I think this is fine, it's weird how even tho you should get the username from there you still use the ID. Can you print the value you get from the URL just to see that it's the username?

lordYorden:
I will check, I think it still is

lordYorden:
![image](https://github.com/user-attachments/assets/8101ae9b-e392-4146-81c3-a10252179657)
it dose show the username

lordYorden:
we should maybe add a DisplayName field.
when @Aradon267 changed his username it showed in the form URI instead
![image](https://github.com/user-attachments/assets/3d9c9287-cac6-45f0-8733-48a4e938dc69)
I added a decode layer to combat this, but we all agree we need a solution @YechiamW3 what do you think?

YechiamW3:
@lordYorden not that I disagree with display name, but in what relation do you bring it up, as related to the profile page?
Let's start with the fact that we should probably disallow special chars for the username (which would eliminate the case you've brought up). That's true either way.
[...]

PR #204, opened 2024-10-31
85 unified editing of data

YechiamW3:
Note: still left to handle username change; I've specifically left it because a small discussion is needed: because we want to (at least) rely on username for the profile page routing, changing it means a different route for the profile - so when the user changes the username we'll need to also redirect to the new page. Do we want to keep previous usernames' pages and redirect via their link to the new username page? I can also just add a cache tag to the previous username's tag to redirect to the new username; but that can be done efficiently only to the last-used username. Or to be more exact, doing it for _all_ the previous usernames will be cumbersome, and I don't think we need to handle such case.

Anyway, redirect to the new username page - a must. Updating the currently previous username's cache to redirect to the new one - bonus. Should I add it?


YechiamW3:
Also @Aradon267, I've added a note on what we discussed regarding validation, let me know if I can safely delete it or keep it for now:
https://github.com/magshimim-next/cv-next/pull/204/files#diff-275c1e4152a47eb79a69accd42fa86187b6dced472f98a5ebce27dca1af59283R91-R97

Aradon267:
I'll remove the constraint and yes this can be removed

Aradon267:
And about what you said about the username: currently, the links are based on getting the path from the URL(the UUID) and then fetching all data based on that. The switch that the URL would have the username is happening on #202 so maybe we can leave it like that and the links will still be based on the usernames?

Aradon267:
Several things that I'm not sure where to put:
1. I needed to run `npm i react-hook-form` first
2. Going on a different profile gave me my data:
image

3. the work status shouldn't be with _ in it, it was like that for the supabase enum

YechiamW3:
> Several things that I'm not sure where to put:
> 
> 1. I needed to run `npm i react-hook-form` first
> 2. Going on a different profile gave me my data:
>    image
> 3. the work status shouldn't be with _ in it, it was like that for the supabase enum

1. You needed to do 'npm i', because a new package was added (and the package.json was modified accordingly). You had to do specifically `npm i react-hook-form` even after doing `npm i`?
2. Welp, will check
3. Right, forgot to check work status. Will fix, thanks

YechiamW3:
> And about what you said about the username: currently, the links are based on getting the path from the URL(the UUID) and then fetching all data based on that. The switch that the URL would have the username is happening on #202 so maybe we can leave it like that and the links will still be based on the usernames?

OK, I'm leaving the TODO so that we won't forget to handle it.

cv_next/app/profile/[profileId]/components/profilePersonalData.tsx
```diff
@@ -0,0 +1,84 @@
...
+    if (userData && user.id === userData.id) {
+      setIsCurrentUser(true);
+    } else {
+      setIsCurrentUser(false);
+    }
+  }, [user.id, userData]);
+
+  const changePfp = () => {
+    //To be implemented
+  };
```
Comment (Aradon267) [inline cv_next/app/profile/[profileId]/components/profilePersonalData.tsx:32]: Maybe there should be like a "Coming soon" thing when the user clicks on the option itself?

Comment (YechiamW3) [reply cv_next/app/profile/[profileId]/components/profilePersonalData.tsx:32]: I do think we need to handle it as part of V1, no?
If not, then sure. lmk.

Comment (YechiamW3) [reply cv_next/app/profile/[profileId]/components/profilePersonalData.tsx:32]: tbf if we don't have it as part of V1 I can simply remove this edit option.

Comment (Aradon267) [reply cv_next/app/profile/[profileId]/components/profilePersonalData.tsx:32]: It's just that currently the way we handle the images is just via the links from gmail(we don't upload any image or something to our own bucket)

Comment (YechiamW3) [reply cv_next/app/profile/[profileId]/components/profilePersonalData.tsx:32]: So, we currently don't have the means to edit the image, correct?
If not, I'll remove the edit option, and we can add it as an issue to V2.

Comment (Aradon267) [reply cv_next/app/profile/[profileId]/components/profilePersonalData.tsx:32]: Exactly, we don't
I'll add it to V2

Issue #125, opened 2024-08-17
UI: append initial commenter's name on new comments


Aradon267:
For now only the display name is shown, maybe in the future a link or something could be added

YechiamW3:
Can we have a bit more details here?

Aradon267:
If I remember correctly this was aout the UI of new comments on inner comments - we decided to leave the inner comments but just add the name of the commenter(if I click on an inner comment that you wrote, my comment will start with your name)

Aradon267:
Maybe we need to wait with this one until #202 is closed

Issue #124, opened 2024-08-17
BUG: new comment re-renders all comments


Aradon267:

YechiamW3:
Can utilize optimistic UI update to solve this (and provide better UX along the way). See e.g. https://swr.vercel.app/docs/mutation#optimistic-updates

Issue #103, opened 2024-08-10
BUG: uploading private CVS


Aradon267:

Aradon267:
@AdamLiberov iirc you said you already finished it, can you create a PR so we could start checking this?

PR #209, opened 2024-11-01
Comments bug fixes&uiux

Aradon267:
This PR combines a lot of issues according to what I know(#103 #123 #125 #124 )
I don't know if this is a draft or not so I'm leaving it as one for now, but I'm making this so we can get started on it


cv_next/app/actions/users/getUser.ts
```diff
@@ -66,6 +66,8 @@ export async function signInWithSocialProvider(provider: any, nextURL: string) {
           nextURL,
       },
     });
+  console.log(data);
+  console.log(error);
```
Aradon267, line 70: Remove these and other console logs as well

cv_next/package-lock.json
YechiamW3, line 1: Those package-lock file changes are fishy, especially when you didn't make any package-related changes.
Make sure these are legit removals, and if not then delete these files, reinstall packages and see if the diff still appears.

package-lock.json
YechiamW3, line 1: Well yeah ok here it should be fine; there shouldn't be any package in this directory.

cv_next/package-lock.json
Aradon267: Took a bit over this, not sure if this change is needed because after I tried to merge from develop I do still see this library and it seems that everything is working fine

YechiamW3 (approved):

cv_next/app/cv/[cvId]/components/commentSection/commentsSection.tsx
```diff
@@ -58,7 +58,7 @@ export default function CommentsSection({ cv }: { cv: CvModel }) {
   }, [pathname, router, supabase.auth]);
 
   return (
-    <>
+    <div style={{ height: "90vh", overflow: "auto" }}>
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/commentsSection.tsx:61]: Missed this; replace `style` with tw classes

Issue #201, opened 2024-10-26
FEATURE: Detect first login


Aradon267:
Find a way to see if it's the user's first login(this means that it's after the user was approved by the mods i.e. it's an active user's first login).

After that we can prompt it o change it's username and in the future add a feature guide or something

lordYorden:
updated_at field should be just created at, it doesn't update with the table in the firebase auth

lordYorden:
as for checking the first login we can use this field to compare it with `now()`

lordYorden:
used `superbase.auth.getUser()` and and created component that check `user.last_sign_in_at == user.created_at` on client side
![WhatsApp Image 2024-10-31 at 17 53 35_5270dbb0](https://github.com/user-attachments/assets/e3bc2620-b634-47a8-86ea-677ae94efca9)

lordYorden:
@Aradon267  I fixed some issues with the tooltip component (added hover and children)
but it caused some weird behavior with comments 
![WhatsApp Image 2024-10-31 at 17 54 10_b61b2ac2](https://github.com/user-attachments/assets/1115b540-51cf-4019-a595-36955cebb8f9)
tried to fix it with no luck

lordYorden:
fixed
![image](https://github.com/user-attachments/assets/f2360b8c-7034-412e-8c29-d927692bc659)

lordYorden:
and this how the sign in tooltip looks
![image](https://github.com/user-attachments/assets/439ba88a-b4e1-4e44-a8a1-70f0b67fd067)

Aradon267:
Won't the created at be affected even if the user wasn't active? 
Other than that, looks good!

lordYorden:
> Won't the created at be affected even if the user wasn't active? Other than that, looks good!

don't know we need to find a viable way to test it, according to documentation it won't

Aradon267:
I'll give it a go with a different Gmail I own

lordYorden:
just change the `line 16` on `firstTimeSignIntsx`

Aradon267:
I logged in and out several times and I don't see the thing there

lordYorden:
you wont see it there except the very first time you log in.
that is way it is very hard to test. we can just detect the hash and prompt it then. 
so it will appear until you update your username.

lordYorden:
could be a better alternative

lordYorden:
```ts
const user = await getUser();
if (!user.ok) return; //TODO: maybe push to inactive page
[...]

PR #218, opened 2024-11-09
Update all usage of user's ID with new username

lordYorden:
as discussed in #202 
updated all user_id instances to use username and full_name
and added validate username flow

## to summarize:
- [x] change the feed function to `getDisplayNameHandler`
- [x] change the field to be `display_name` instead of `full_name`
- [x] change variable names in `cvData` to fit the new flow
- [x] cleanup logs and comments 
- [x] change profile keys as well as field name in the db

@YechiamW3 please confirm that is all


cv_next/app/api/users/fetchUserData/route.ts
```diff
@@ -32,7 +32,7 @@ async function getUserNameHandler(data: { userId: string }) {
       { status: 500 }
     );
   } else {
-    const userName = result.val.username ?? result.val.full_name;
+    const userName = result.val.full_name || result.val.username;
```
YechiamW3, line 35: Wait, is the username here supposed to be display name?
Because if it's username, then we should prefer `username` as default and `full_name` as fallback (well, tbh we shouldn't have a fallback as it always should have a value).
If it's the display name, make sure that it's true throughout the usage of this route.

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -387,9 +387,9 @@ export default function Comment({
           <p className="mr-3 inline-flex items-center text-sm font-semibold text-gray-900 dark:text-white">
             <Link
               className="text-lg font-medium hover:underline"
-              href={`/profile/${commenter.id}`}
+              href={`/profile/${commenter.username}`}
             >
-              {commenter.username || commenter.full_name}
+              {commenter.full_name}
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:392]: Can we also change the field name to `display_name` instead of `full_name`? It'll be more consistent with what we're trying to convey with this field.

cv_next/app/cv/[cvId]/components/cvData.tsx
```diff
@@ -7,9 +7,7 @@ import DynamicProfileImage from "@/components/ui/DynamicProfileImage";
 
 export default function CvData({ cv }: { cv: CvModel }) {
   const uploader = JSON.parse(JSON.stringify(cv.user_id));
-  const userName = uploader.username?.length
-    ? uploader.username
-    : uploader.full_name;
+  const userName = uploader.full_name || uploader.username;
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/cvData.tsx:10]: So, again - `full_name` (or display name) doesn't have any API meaning, so hrefing to `/profile/full_name` is meaningless. You should take *only* `uploader.username` - if it doesn't exists then it's an error!

cv_next/app/api/users/fetchUserData/route.ts
lordYorden: We decide to use the display name for client side username and username as a replacement for user id. You are right the function should be called get display name instead

cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const ProfileDisplay = ({ user }: { user: UserModel }) => {
+  return (
+    <>
+      <div className="m-auto w-full">
+        <div className="flex flex-wrap justify-between">
+          <label className="font-bold" htmlFor="username">
+            Username:{" "}
+          </label>
+          <span id="username">{user.full_name}</span>
```
Comment (YechiamW3) [inline cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: Now you should have two field displays - both the username and the display name. See e.g. github:
image

cv_next/app/profile/[profileUsername]/page.tsx
```diff
@@ -5,22 +5,26 @@ import { unstable_cache } from "next/cache";
...
 
 export default async function Page({
   params,
 }: {
-  params: { profileId: string };
+  params: { profileUsername: string };
 }) {
-  const { profileId } = params;
+  const { profileUsername } = params;
+  logger.info("profileUsername: " + profileUsername);
```
Comment (YechiamW3) [inline cv_next/app/profile/[profileUsername]/page.tsx:18]: Don't forget to delete the log

cv_next/app/profile/[profileUsername]/page.tsx
```diff
@@ -5,22 +5,26 @@ import { unstable_cache } from "next/cache";
...
+  const cleanUsername = decodeURIComponent(profileUsername);
 
-  const userFetcher = getUserModel(profileId);
+  const userFetcher = await getUserModel(cleanUsername);
+  //const userFetcher = getUserModel(profileId);
```
Comment (YechiamW3) [inline cv_next/app/profile/[profileUsername]/page.tsx:22]: And the dangling comment

cv_next/app/cv/[cvId]/components/cvData.tsx
Comment (lordYorden) [reply cv_next/app/cv/[cvId]/components/cvData.tsx:10]: Read the code again before you assume the uses of a variable 
I specified the use of username in the href so it is only for display purposes

cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx
Comment (lordYorden) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: That is not what we talked about, at this point we use username as a client replacement to uuid so it remains
constant as generated for the time being. If we would like to implement such a change, it would be on the next version. if and when we do a db overhall

cv_next/app/profile/[profileUsername]/page.tsx
Comment (lordYorden) [reply cv_next/app/profile/[profileUsername]/page.tsx:18]: my mistake

cv_next/app/profile/[profileUsername]/page.tsx
Comment (lordYorden) [reply cv_next/app/profile/[profileUsername]/page.tsx:22]: that was a merge cleanup with develop

cv_next/app/cv/[cvId]/components/cvData.tsx
Comment (lordYorden) [reply cv_next/app/cv/[cvId]/components/cvData.tsx:10]: you are right that we need to check uploder.username exist

Comment (YechiamW3) [reply cv_next/app/cv/[cvId]/components/cvData.tsx:10]: > Read the code again before you assume the uses of a variable I specified the use of username in the href so it is only for display purposes

Correct - then the variable should be easier to distinguish (e.g. `displayName`, not `userName`).

cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx
Comment (YechiamW3) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: Not sure I understand.
Username is a unique id that can be used for client-side usage instead of the uuid.
The display name/full_name is the displayable name of the user (to be used in comments, cvs etc).
What is there to implement further (aside from what you've already done)?

Comment (lordYorden) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: so let me understand you want to also display on profile the generated username?

Comment (lordYorden) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: instead of displaying and editing the displayname only

Comment (YechiamW3) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: OK, two points:
1. Without regarding point 2, the full name is display name, and not username. Let's keep this terminology throughout the project. So if you're displaying the full name (=display name), the label should be "display name".
2. Remember that we require the user to change the generated username - it's only generated for first usage. Perhaps I've missed a discussion, but from what you're saying it seems that you've decided to only let the user change their display name, and not their username? If so, why? In most sites where you have both display name and username, you can change both attributes..

Comment (lordYorden) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: in #202 we agreed to make only the display name editable in order to use the generated username to use as replacement for user id in client side. And so it will not change instead there is a new flow in `app/callback`
we use `validateUsername` to regenerate if its null and when we generate it we also change the display name temporarily to fit. and the user is able to change the display name

Comment (lordYorden) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: maybe there was a miscommunication but I showed it in the weekly meeting and @Aradon267 was pleased

Comment (YechiamW3) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: OK I'm sorry I might have not explained myself properly.
What I meant is the following:
We agreed on having a display name property (that I said we don't need to be a new one, but simply a rename of full_name). When I said to make it editable I didn't mean *instead* of the username, but rather *in addition to*, because username, I believe, should stay editable. It's still a visible property of the user, so the user should be able to edit it as they please (as long as it stays unique).
@Aradon267 additional thoughts?

Comment (lordYorden) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: ok so I understand the confusion, and that fine by me. but just so were on the same page it will cause `URI problems` because we it to fetch the user in profile

Comment (lordYorden) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx:12]: also it means adding a new field in the unified editing of profile data

Comment (YechiamW3) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx]: Adding a new field - I agree. But it shouldn't be more than 1-2 lines of codes iirc.
Causing URI problems - I agree as well, that's why I left a comment on the username update that it'll require a bit more complicated logic than simple update.
Honestly, I'm not 100% against removing that option - but I believe that in every site where you have a username, you generally also have an option to edit it. So it makes sense that users will ask for it.
And frankly, we use the user's id in all the server logic, so changing the username won't cause any issue other than the profile page. If you want, I can sketch a solution.

Comment (lordYorden) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx]: I tell you what if we were to show the display name as the username wouldn't that will solve your problem?

It's like the old discord model you have your 'discord tag' and the username that you are able to change (In our case the 'display_name' is the username, and the tag is the 'generated username'). We could even go as far as regenerating the tag when the 'display_name' changes.

Sounds good?

Comment (Aradon267) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx]: Just to make sure that I understood @lordYorden the discord thing will leave the tag(the generated field) uneditable by the user if it's not via regenerating based on display name changes? I don't know If that's preferable than the username and display name being changeable(based on experience with for example how LinkedIn handles their user links)

Comment (YechiamW3) [reply cv_next/app/profile/[profileUsername]/components/profileDisplay.tsx]: I agree with Arad, also the issue with relying on display name for basically anything other than profile page (not saying that it isn't useful there - it's very useful there, especially in this kind of project) is distinguishability. Not saying that there's a high chance for it to happen - but the chance 100% exists, that in a single comments section we'll have two people with the same name.

YechiamW3:
To summarize the discussion from Friday:
- change full_name field to display name (db, code)
- populate display name with full_name on first authentication (already happening)
- on first login (recognized via new boolean field in db), show popup to change username from generated one (validate that it's actually different, and still unique) - will be done as part of #201
   - optional: add option to change the display name there - but don't make it required (only the username)
- show username in the profile page (and allow change if it's your page)
   - optional: show username only if it's your page

I think that's it, but comment if I missed anything.

lordYorden:
image

it seems to be working

cv_next/server/api/users.ts
```diff
@@ -382,3 +375,33 @@ export async function setDisplayName(
...
+): Promise<Result<void, string>> {
+  const id = await getCurrentId();
+  if (!id.ok || !id.val) {
+    return Err(setFirstLogin.name, { err: Error("No user was connected!") });
+  }
+
+  try {
+    const { error } =
+      await SupabaseHelper.getSupabaseInstance().auth.updateUser({
+        data: { is_first_login: isFirstLogin },
```
YechiamW3, line 396: Not entirely sure of the difference, perhaps @Aradon267 knows, but why `updateUser` and not regular table update (like the other functions in this file)?

Aradon267: UpdateUser updates the authentication metadata from the 3rd party provider(auth.users) unlike the other that updates our data(public.profiles)

YechiamW3: Is it needed here? 🤔

lordYorden: yes, that is what we agreed in #201.
This is meant to be the setup for the `first_login` state

lordYorden:
> * on first login (recognized via new boolean field in db), show popup to change username from generated one (validate that it's actually different, and still unique) - will be done as part of [Detect first login #201](https://github.com/magshimim-next/cv-next/issues/201)

also mentioned here

YechiamW3 (approved):

PR #220, opened 2024-11-21
219 UI remove errors on upload page

Aradon267:
1. The user will now get error signs only after interacting with the input fields.
2. The user must choose at least 1 category.
3. "category" has been fixed to "category"


cv_next/app/upload/page.tsx
```diff
@@ -48,24 +48,28 @@ const InputRow = ({
...
 export default function Page() {
-  const [catagoryId, setCatagoryId] = useState<number[]>([]);
+  const [categoryId, setCategoryId] = useState<number[]>([]);
   const [description, setDescription] = useState("");
   const [link, setLink] = useState("");
   const { showError } = useError();
 
+  const [isLinkTouched, setIsLinkTouched] = useState(false);
+  const [isDescriptionTouched, setIsDescriptionTouched] = useState(false);
+  const [isCategoryTouched, setIsCategoryTouched] = useState(false);
```
YechiamW3, line 58: It might be easier to use `react-hook-form`, I already started incorporating it into our codebase in #204. It gives you these kinds of options as well (know when the field has been touched/"dirtied").

Aradon267: Updated this

cv_next/app/feed/components/filters/valueSelect.tsx
```diff
@@ -68,9 +65,18 @@ export const DropdownInput = ({
...
+    }
     setSelectTitle(selectedCategories.map(getValueById).join(", "));
-  }, [selectedCategories, onChange, setSelectTitle, getValueById]);
+  }, [selectedCategories, valueId, getValueById, onChange]);
+
+  useEffect(() => {
+    if (JSON.stringify(valueId) !== JSON.stringify(selectedCategories)) {
+      setSelectedCategories(valueId || []);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
YechiamW3, line 78: why? disabling exhaustive deps should be used in rare cases only

Aradon267: selectedCategories changes on every render after something is selected as handleCheckboxChange occurs when we select something, and it updates selectedCategories. This useEffect also uses setSelectedCategories so it's a loop... I'll try and combine both things and see what I can do

Aradon267: Managed to fix this issue without another useEffect thus removing the exhaustive deps

lordYorden (changes requested):
image
@Aradon267 error goes away when selecting undefined (shouldn't even be an option)

Aradon267:
@lordYorden I've removed the undefined option from the dropdown

lordYorden (commented):
image
@Aradon267 categories seems to be fixed but the cv preview has a weird middle frame

Aradon267:
By middle frame you mean this?
image

This is how the iframe works with the docs when using some URL parameters like minimal etc

lordYorden (approved):

PR #229, opened 2024-12-13
Comment rendering fixes

Aradon267:
This PR fixes:
1. New comments don't rerender other comments.
2. Starting new comments on comments appends the parent commenter's display name:
image


Aradon267:
There's a problem where when someone changes the comment it isn't changed

Aradon267:
> There's a problem where when someone changes the comment it isn't changed

This has been fixed.

Also I added an at-sign(@) like this:
image

YechiamW3:
> > There's a problem where when someone changes the comment it isn't changed
> 
> This has been fixed.
> 
> Also I added an at-sign(@) like this: image

This is great, but I assume it currently doesn't trigger any functionality right?

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -281,62 +298,32 @@ export default function Comment({
...
+      {
+        ...commentToAdd,
+        id: Date.now().toString(),
+        deleted: false,
+        last_update: new Date().toISOString(),
+        resolved: false,
+        upvoted: [],
+        user_id: userId,
+      },
+    ]);
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:312]: We don't want optimistic behavior?

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -281,62 +298,32 @@ export default function Comment({
...
-        .get(parentId)!
-        .push(commentToAdd as unknown as CommentModel);
-
-      return Array.from(newCommentsOfComments.values()).flat();
-    };
+    mutate(comment.document_id, (currentSubComments: CommenterModel[] = []) => [
+      ...currentSubComments,
+      {
+        ...commentToAdd,
+        id: Date.now().toString(),
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:305]: Also- what? Why define id as date.now()? Why not let the db create the id?

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -281,62 +298,32 @@ export default function Comment({
...
-            if (error instanceof Error) {
-              return error.name !== "AbortError";
-            }
-            return true;
-          },
-          revalidate: true,
-        }
+        (currentComments: CommentModel[] = []) =>
+          currentComments.filter((comment) => comment.id !== comment.id),
+        false
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:323]: You don't really need to rollback if you don't use optimistic behavior

Aradon267:
> > > There's a problem where when someone changes the comment it isn't changed
> > 
> > 
> > This has been fixed.
> > Also I added an at-sign(@) like this: image
> 
> This is great, but I assume it currently doesn't trigger any functionality right?

That's correct

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -24,14 +23,29 @@ interface NewCommentBlockProps {
...
+  parentCommenter,
 }: NewCommentBlockProps) => {
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [inputValue, setInputValue] = useState(`@${parentCommenter} `);
+
+  useEffect(() => {
+    if (commentOnCommentStatus && inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [commentOnCommentStatus]);
```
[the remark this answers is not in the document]
Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:43]: I'm not getting any warning here and when trying to add it anyway:
```
React Hook useEffect has an unnecessary dependency: 'inputRef.current'. Either exclude it or remove the dependency array. Mutable values like 'inputRef.current' aren't valid dependencies because mutating them doesn't re-render the component.
```

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:305]: It's a temporary thing until we load the actual comment that we get from the server(which is done immediately after although we don't trigger re-render this way)

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:312]: Isn't that considered optimistic behavior? Adding a new comment locally because we believe that the server would add it?

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:323]: See my reply above about this, if we decide that it isn't related I'll remove this

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (YechiamW3) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:312]: Commenting only here: what I don't understand is, why aren't you using the `optimisticData` and `rollbackOnError` props of mutate?

Comment (YechiamW3) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:312]: (e.g. as it were before: `mutate(comment.document_id, mutationFn, { optimisticData: withNewData, rollbackOnError: (conditional) true });`)

Comment (YechiamW3) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:312]: See here for example: https://swr.vercel.app/examples/optimistic-ui

Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:312]: The main difference is that their API returns all of the objects after adding unlike ours which returns nothing(and even if we do return the object created it still doesn't work as intended due to type mismatches...)

Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx]: Ok, I managed to come up with a solution that actually uses optimistic UI

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx]: Removed this from the code

lordYorden:
works pretty well, do you want to add `enter key callback` as an alternative for the + button?
just feels better when typing.

Aradon267:
enter key as in enter to send? I can I'll just have to add a ctrl enter handler

Aradon267:
@lordYorden I added the ctrl and enter functionality

lordYorden (approved):

PR #244, opened 2024-12-28
243 bug popup information not updating in time after redirect and db update

Aradon267:
1. mutateUser is called after updating the username
2. display_name can be username(value-wise)


lordYorden (approved):
seems ok, tested display name and first login

PR #245, opened 2024-12-28
240 UI cv statistic get cut out on smaller screen

Aradon267:
On laptops and similar:
image

On phones:
image

On tablets:
image


lordYorden (approved):
looks ok

PR #268, opened 2025-01-13
201 Detected first login and ask the user to change the default username that is generated

lordYorden:
## includes (as discussed at #201)
- [x] new page called first_login 
- [x] new user state in `auth` db scheme named: `isFirstLogin`
- [x] automatically transfer the user when is in state in `auth/callback.tsx`
- [x] a fix for tooltips used in `components\firstTimeSignIn.tsx`

### To summarized (things to fix):
- [x] regex check on server side (add dot support)
- [x] inform the users on error (duplicate username) 
- [x] feedback for not passing validation
- [x] maximum length for username
- [x] fix confirm button clickable even if disabled


cv_next/app/auth/callback/route.ts
```diff
@@ -4,24 +4,31 @@ import Definitions from "@/lib/definitions";
...
+            validateUsername.name,
+            "Username was generated successfully: " + isValid.val
+          );
+
+          const isFirstLogin = await isCurrentFirstLogin();
+          if (isFirstLogin.ok) {
+            next = `${Definitions.FIRST_LOGIN_REDIRECT}/${isValid.val}`;
+            // return NextResponse.redirect(
+            //   `${process.env.NEXT_PUBLIC_BASE_URL}${Definitions.AUTH_DEFAULT_REDIRECT}${Definitions.FIRST_LOGIN_REDIRECT}${Definitions.AUTH_DEFAULT_REDIRECT}${isValid.val}`
+            // );
```
Aradon267, line 29: This can be removed

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
```diff
@@ -149,16 +153,17 @@ const CommenterActions = ({
...
-            <FaRegTrashCan
-              data-tooltip-target="Trash Icon"
-              style={{
-                transform: "translateY(2px)",
-              }}
-              fontSize="1.4rem"
-            />
-          </button>
-          <Tooltip id="Trash icon" message="delete"></Tooltip>
+          <Tooltip id="Trash icon" message="delete">
```
Comment (Aradon267) [inline cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:156]: Capitalize

cv_next/app/first_login/[profileUsername]/components/firstTimeSignIn.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const [isFirstLogin, setIsFirstLogin] = useState<Boolean>(false);
+  const superbase = createClientComponent();
+
+  useEffect(() => {
+    async function checkFirstLogin() {
+      const res = await getFirstTimeLogin();
+      if (res.ok) {
+        setIsFirstLogin(res.val);
+      } else {
+        //TODO: push to inactive page
```
Comment (Aradon267) [inline cv_next/app/first_login/[profileUsername]/components/firstTimeSignIn.tsx:18]: Add this with next navigation

cv_next/app/first_login/[profileUsername]/components/firstTimeSignIn.tsx
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    checkFirstLogin();
+  }, [superbase.auth]);
+
+  return (
+    <>
+      {isFirstLogin && (
+        <Tooltip
+          message="First time sign in? Change your username"
```
Comment (Aradon267) [inline cv_next/app/first_login/[profileUsername]/components/firstTimeSignIn.tsx:29]: fix to signing in

Aradon267 (changes requested):
Mainly semantics and stuff
Also make sure that you add this route to the protected routes in middleware and to the routes that are allowed to be sent with next

cv_next/app/first_login/[profileUsername]/components/firstTimeSignIn.tsx
Comment (lordYorden) [reply cv_next/app/first_login/[profileUsername]/components/firstTimeSignIn.tsx:18]: I don't think it is needed anymore forgot to remove it

cv_next/app/cv/[cvId]/components/commentSection/comment.tsx
Comment (lordYorden) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx:156]: you mean `Trash Icon`?

Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/comment.tsx]: Yes

cv_next/app/first_login/[profileUsername]/page.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const current = await getCurrentId();
+  const isFirstLogin = await isCurrentFirstLogin();
+
+  if (
+    res === null ||
+    !current.ok ||
+    !res.ok ||
+    res.val.id !== current.val ||
+    !isFirstLogin.ok ||
+    !isFirstLogin.val
```
Comment (Aradon267) [inline cv_next/app/first_login/[profileUsername]/page.tsx:24]: If we couldn't get the user based on the name or if we couldn't get the current ID or if the ID and the supplied username mismatch or it's not the first login or we couldn't get the value of first login? because when I print isFirstLogin on a new user I get  `{ ok: true, val: undefined }`

Aradon267 (changes requested):

cv_next/lib/definitions.ts
```diff
@@ -13,6 +13,7 @@ export default class Definitions {
   public static readonly FETCH_WAIT_TIME = 120;
   public static readonly COMMENTS_REVALIDATE_TIME_IN_SECONDS = 0;
   public static readonly AUTH_CALLBACK_REDIRECT = "/auth/callback";
+  public static readonly FIRST_LOGIN_REDIRECT = "first_login";
```
Aradon267, line 16: Also add this to allowed redirects in definitions

lordYorden: @Aradon267  i didn't understand it is already there. what do you mean by add this

cv_next/app/auth/callback/route.ts
```diff
@@ -4,24 +4,28 @@ import Definitions from "@/lib/definitions";
...
 export async function GET(request: Request) {
   const { searchParams, origin: _origin } = new URL(request.url);
   const code = searchParams.get("code");
-  const next = searchParams.get("next") || "";
+  let next = searchParams.get("next") || "";
   if (code) {
     const supabase = SupabaseHelper.getSupabaseInstance();
     const { error } = await supabase.auth.exchangeCodeForSession(code);
     if (!error) {
       if (checkRedirect(next)) {
```
Aradon267, line 16: Make sure that the checkRedirect accepts the path as well

cv_next/lib/definitions.ts
Aradon267: Added in a more relevant place

cv_next/app/first_login/[profileUsername]/page.tsx
Comment (lordYorden) [reply cv_next/app/first_login/[profileUsername]/page.tsx:24]: didn't understand the concern the code behind this is the following:
```ts
return Ok(metadata.is_first_login); 
```

it just return the the value in the metadata of auth.getUser();
@Aradon267 what user did you try this on and did you check `supabase` to see if he entered the `isFirstLogin = true` state

Comment (Aradon267) [reply cv_next/app/first_login/[profileUsername]/page.tsx:24]: I signed in with a brand-new user

Comment (lordYorden) [reply cv_next/app/first_login/[profileUsername]/page.tsx:24]: ### to explain this if briefly - we don't want to show the page if:
1. The username requested is not the same as the one logged in to the site
2. There is no username logged in to the site
3. The user is correct and matches the one logged in but, That user is not in the `isFirstLogin = true` state

cv_next/app/auth/callback/route.ts
lordYorden: @Aradon267 That I don't quite understand we filter client redirects but why check a constant and internal const
```ts
const isFirstLogin = await isCurrentFirstLogin();
if (isFirstLogin.ok) {
  next = `${Definitions.FIRST_LOGIN_REDIRECT}/${isValid.val}`;
}

```
it is the only time that value changes.

Aradon267: Ohhh ok

cv_next/app/first_login/[profileUsername]/page.tsx
Comment (Aradon267) [reply cv_next/app/first_login/[profileUsername]/page.tsx:24]: new user, got inactive and disconnected
updated to active and reconnect
got a generated username
is first login = true
getting 404 although I am redirected to the page

Comment (Aradon267) [reply cv_next/app/first_login/[profileUsername]/page.tsx:24]: Ok it randomly worked after restarting npm

cv_next/app/first_login/[profileUsername]/components/newUsernameForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+            />
+          </div>
+        </div>
+
+        <div className="mt-5 flex max-w-sm items-center justify-center">
+          <Button
+            text="Confirm"
+            onClick={onConfirm}
+            isDisabled={!validate.newUsername()}
+          />
```
Comment (Aradon267) [inline cv_next/app/first_login/[profileUsername]/components/newUsernameForm.tsx:86]: For some reason, although I used space, it was still clickable(it was seen as disabled, but still clickable)
So this and adding a validation on the server is also important

cv_next/app/first_login/[profileUsername]/components/newUsernameForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+      await setFirstLoginCurrent(false);
+      await mutateUser();
+      router?.push(`/profile/${newUsername}`);
+    }
+  }
+
+  const validate = (() => {
+    const checkNewUsername = () => {
+      //only alphanumeric characters and underscores
+      const regex = /^[a-zA-Z0-9_]+$/;
```
Comment (Aradon267) [inline cv_next/app/first_login/[profileUsername]/components/newUsernameForm.tsx:38]: Not regex related but do we also want a minimum length?

cv_next/app/first_login/[profileUsername]/components/newUsernameForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const router = useRouter();
+
+  async function onConfirm() {
+    const res = await getUserModel(profileUsername);
+
+    if (res === null || !res.ok) {
+      return;
+    }
+
+    const isValid = await setNewUsername(res.val.id, newUsername);
```
Comment (Aradon267) [inline cv_next/app/first_login/[profileUsername]/components/newUsernameForm.tsx:27]: I think that there should be some user feedback about this(mainly if the error is `duplicate key value violates unique constraint` just to tell them that a user with this username already exists(useError is fine for that)

cv_next/app/first_login/[profileUsername]/components/newUsernameForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+    if (isValid.ok) {
+      await setFirstLoginCurrent(false);
+      await mutateUser();
+      router?.push(`/profile/${newUsername}`);
+    }
+  }
+
+  const validate = (() => {
```
Comment (Aradon267) [inline cv_next/app/first_login/[profileUsername]/components/newUsernameForm.tsx:35]: Maybe there should be a small text with the problem like the thing in upload(the username can only contain letters, numbers, and underscore type)
This can be done using react jook forms which might be good to incorporate here

Aradon267 (changes requested):

Aradon267:
Some other PT things that I noticed that we need to think about:
1. While we can't do multiple usernames, do we care if the username Araddo exists and then another user changes to araddo(small a)?
2. Should we also allow dots in the regex? I mean if the generated name contains them, is it a problem?
3. Do you know why I get this warning sometimes?
```
Using the user object as returned from supabase.auth.getSession() or from some supabase.auth.onAuthStateChange() events could be insecure! This value comes directly from the storage medium (usually cookies on the server) and many not be authentic. Use supabase.auth.getUser() instead which authenticates the data by contacting the Supabase Auth server.
```
Don't even see it in the code

lordYorden:
### To summarized (things to fix):

- [ ] regex check on server side (add dot support)
- [ ] inform the users on error (duplicate username) 
- [ ] feedback for not passing validation

@Aradon267 did I understand you correctly?

### As for your questions:
1. The username is `case-sensitive` which means we don't care for variations of the same name.
2.  The regex account for minimum length using "+" which means one or more (we can change the regex to fix this if you want)
3. I did not encounter that error, were did it showed (which page and which function/line num)?

Aradon267:
Regex check on server + add dot + if possible check what's wrong with client
Th minimum is fine but we should also add a maximum
The error is a console.error(or similar) that happens straight in the supabase function itself so not something we printed(meaning I have no idea where is that)

lordYorden:
1. got it
2. no problem, What do you think the limit should be?
3. so the error is for superbase internally?
4. why not `case-sensitive` it is very common on apps these days?

Aradon267:
> 2. no problem, What do you think the limit should be?
> 

Maybe 20/30? What do we have in the display name?

> 3. so the error is for superbase internally?
> 

Yes it's an error that is shown to alert the user that he's using something unsafe

> 4. why not `case-sensitive`? it is very common on apps these days

Won't it cause fake people or something?

lordYorden:
1. not sure we have it on the `display name` either, Do you mean on DB?
2. that's ok so it isn't out fault. maybe there is a deprecated function (maybe we need to update superbase dependency)
3. that's valid, but not a true concern for now (You can open an issue if you want to).

Aradon267:
There might be something about it in the DB constraints

lordYorden:
image
added a `useError for duplicate username`

Aradon267:
> image
> 
> added a `useError` for duplicate username

Good
Haven't looked at the code but make sure that it's in the definitions(there's a place for these things)

lordYorden:
> > image
> > added a `useError` for duplicate username
> 
> Good Haven't looked at the code but make sure that it's in the definitions(there's a place for these things)

already did

lordYorden:
> There might be something about it in the DB constraints

image

both are `text` so there is no max characters currently

lordYorden:
image
added validation msg

lordYorden:
@Aradon267 fixed all that you have asked

Aradon267:
> @Aradon267 fixed all that you have asked

out of curiosity, how did you fix the client button?

Aradon267 (approved):

lordYorden:
> > @Aradon267 fixed all that you have asked
> 
> out of curiosity, how did you fix the client button?

@Aradon267 fixed it now

lordYorden:
tested the entire process and merge it when you see this

PR #281, opened 2025-01-19
Removed navigation buttons from navbar

TheGreenOak:
- [x] PC
- [x] Mobile


Aradon267 (commented):
This is how it looks without any buttons:
image

1. Isn't it a bit empty?
2. Should we add the about page and hall of fame on the footer?
3. This is how it looks on mobile
image
Is that good?

Aradon267:
Updated it so that on mobile it goes straight to the bigger profile component
image

lordYorden (approved):
looks great @Aradon267 
agree about the footer for hall of fame

Aradon267:
> looks great @Aradon267 agree about the footer for hall of fame

I'll do that in the update in #297

PR #284, opened 2025-01-24
CHORE: 276 combine login and about and minimize about page

lordYorden:
- [x] need to remove `\about`


Aradon267 (changes requested):
Should the signout also have Google's logo?
image

Should we add something that signals users to signing if they click the feed for example?

lordYorden:
@Aradon267 @TheGreenOak we can maybe move the `/about` to /login instead
what do you think?

lordYorden:
> Should the signout also have Google's logo? image
> 
> Should we add something that signals users to signing if they click the feed for example?

will fix

lordYorden:
image
@Aradon267 better?

Aradon267:
> image @Aradon267 better?

yes
login is the same right?

lordYorden:
image

Aradon267 (approved):

Aradon267 (approved):

PR #326, opened 2025-03-16
319 bug change comments limitations

Aradon267:
New state:
Commenting when above 750 chars isn't possible.
Commenting when above 750 chars shows a warning(like that of the delete).
Failing to add a new comment doesn't delete the comment we tried to enter from the input area.


lordYorden (changes requested):
@Aradon267 make it so that you can't type beyond the set limit
in both upload and comment

cv_next/app/cv/[cvId]/components/commentSection/commentForm.tsx
```diff
@@ -62,28 +66,34 @@ export default function CommentForm({ cv }: { cv: CvModel }) {
...
-      setText("");
+      const res = await formAction();
+      if (res) setText("");
     }
   };
 
   const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
-    const value = e.target.value;
-    setText(value);
+    if (e.target.value.length <= 750) {
```
Comment (YechiamWe) [inline cv_next/app/cv/[cvId]/components/commentSection/commentForm.tsx:89]: OK sorry used in multiple files; global constant.

Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/commentForm.tsx:89]: Write, updated it

Comment (Aradon267) [reply cv_next/app/cv/[cvId]/components/commentSection/commentForm.tsx:89]: right, updated it

lordYorden (approved):
@Aradon267  add counter and notify the user when the limit is reached
open another issue

Aradon267:
@lordYorden Issue #331 has been opened for the counter
Fixes the issue that it's not showing the error when trying to put more than 750(it still allows it to be submitted because its still 750, but the data doesn't get added)

PR #347, opened 2025-06-01
235 add a download button to the cvs

### LINKED ISSUE #235: Add a download button to the CV's
If someone thinks a cv looks good and wants to copy it or for any other reason wants do download a cv, (and if the download doesn't take too much AWS resources) It should be possible to download a CV.
(make the option to download as PDF and DOC)

Aradon267:
## Description

This PR adds download buttons on CVs. This allows users to obtain a local copy of the CV if they wish to replicate it.
image
image
The difference is due to the usage of Drive and Docs extensions.
White mode:
image
On a phone:
image
image

Fixes #235 

## Type of change

Please delete options that are not relevant.

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Tested using Drive and Docs, with different color schemes and screen sizes

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] My changes generate no new warnings


cv_next/app/cv/[cvId]/components/downloadButtons.tsx
```diff
@@ -0,0 +1,54 @@
+import { PiFilePdfDuotone } from "react-icons/pi";
```
Comment (YechiamW3) [inline cv_next/app/cv/[cvId]/components/downloadButtons.tsx:1]: please add use client

Aradon267:
> approving in principle; probably needs to be adjusted according to discussion on #235

Repeating what I said in the issue itself, I can add this in the management options but does it matter if it's still retrievable from inspect?

YechiamW3:
> > approving in principle; probably needs to be adjusted according to discussion on #235
> 
> Repeating what I said in the issue itself, I can add this in the management options but does it matter if it's still retrievable from inspect?

I see what you mean. Well, we need to consider: do we care about privacy/ownership in these regards, or not. If we care, then we should provide a deeper solution (obfuscate the url or smth, keeping the link data only in the db and server-load the picture); if we don't care, then the feature is good to go as it is.

YechiamW3:
OK. I think perhaps discussing this with Alik - to see what the community thinks about this - would be a good move, but considering we can't contain the privacy issue here, perhaps adding a consent of "you agree to share this cv" or smth when uploading a cv would solve the concern and that's it.

Aradon267:
I'll talk with Alik but I also added a checkbox component and added it to the upload page

YechiamW3 (approved):

Issue #380, opened 2025-09-23
[FEATURE] - create a script to handle previews


Aradon267:
## Is your feature request related to a problem? Please describe

The feed and CV loading take a lot of time. By moving this operation to a standalone script, we could ease that process.

## Describe the solution you'd like

A script that replaces the revalidatePreview flow in the website itself and runs independently.

## Additional context

The script will be created in a different repo

PR #381, opened 2025-10-10
291 chore update cv fetching to use the script

### LINKED ISSUE #291: CHORE: update CV fetching to use the script

Aradon267:
## Description

This PR removes the revalidate functionality from CV previews. It makes it so that the backend just gets the supabase URL and only reuploads it if it's not in Supabase(which also handles if the CV was just uploaded).

Fixes #291 when used with #380 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

Using existing and new CVs with different visibility.

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules


cv_next/app/api/cvs/fetchPreviews/route.ts
```diff
@@ -44,6 +41,20 @@ async function revalidatePreviewHandler(data: {
 
   const docsID = getIdFromLink(cvLink);
   const fileName = docsID + ".png";
+  const publicUrl = SupabaseHelper.getSupabaseInstance()
+    .storage.from(Storage.cvs)
+    .getPublicUrl(fileName).data.publicUrl;
```
lordYorden: `publicUrl` could be undefined

cv_next/app/api/cvs/fetchPreviews/route.ts
```diff
@@ -44,6 +41,20 @@ async function revalidatePreviewHandler(data: {
...
+
+  const supabasePreviewResponse = await fetch(publicUrl, {
```
lordYorden: fetch could fail needs try catch

lordYorden: also better to extract the required output, for exmple:
```ts 
const {status, err} = await fetch(...);
```

lordYorden (changes requested):
@Aradon267 need some error handeling

cv_next/app/api/cvs/fetchPreviews/route.ts
Aradon267: You'd be surprised, but it can't
Apparently, this is just a convenience function from Supabase that just concatenates relevant fields without actually validating anything - that's also why I check if the file exists in supabase or doesn't with the fetch afterward. I will add a comment to make it clear

cv_next/app/api/cvs/fetchPreviews/route.ts
Aradon267: Are you sure that it will throw an error for a simple 404? I'll add a try catch but I don't think that this happens

Aradon267:
@lordYorden about your point from earlier, just note that because we merged the signed previews change, we now don't need to check with try catch, because unlike the public URL, the signed preview does actually check for this

lordYorden (approved):
@Aradon267 got it

